PR #2, opened 2020-08-19
[FLINK-19003][checkpointing] Add micro-benchmark for unaligned checkpoints

zhijiangW:
The results as below:

# Run complete. Total time: 00:12:56

Benchmark                                                                                                                 Mode  Cnt     Score     Error   Units
UnalignedCheckpointBenchmark.unalignedCheckpointWithLocalChannel      thrpt   30  1985.953 ± 156.639  ops/ms
UnalignedCheckpointBenchmark.unalignedCheckpointWithRemoteChannel  thrpt   30  1327.177 ±  86.421  ops/ms

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
@@ -29,6 +29,7 @@ import org.apache.flink.streaming.api.graph.StreamingJobGraphGenerator;
 
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.TearDown;
 import org.openjdk.jmh.runner.Runner;
@@ -47,6 +48,9 @@ public class RemoteChannelThroughputBenchmark extends BenchmarkBase {
     private static final long CHECKPOINT_INTERVAL_MS = 100;
 
     private MiniCluster miniCluster;
+
+    @Param({"ALIGNED", "UNALIGNED"})
+    public String mode = "ALIGNED";
 
     public static void main(String[] args)
             throws RunnerException {
@@ -80,6 +84,7 @@ public class RemoteChannelThroughputBenchmark extends BenchmarkBase {
         StreamExecutionEnvironment env = context.env;
         env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
         env.setParallelism(PARALLELISM);
+        env.getCheckpointConfig().enableUnalignedCheckpoints(!mode.equals("ALIGNED"));
 
         DataStreamSource<Long> source = env.addSource(new LongSource(RECORDS_PER_SUBTASK));
         source
--- a/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.configuration.ConfigConstants;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.TaskManagerOptions;
+import org.apache.flink.runtime.state.CheckpointListener;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.api.functions.sink.SinkFunction;
+import org.apache.flink.streaming.api.functions.source.RichParallelSourceFunction;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.io.IOException;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+/**
+ * The benchmark for measuring the time taken to finish the configured number of
+ * unaligned checkpoints.
+ */
+@OutputTimeUnit(MINUTES)
+@OperationsPerInvocation(value = UnalignedCheckpointTimeBenchmark.NUM_FINISHED_CHECKPOINTS)
+public class UnalignedCheckpointTimeBenchmark extends BenchmarkBase {
+    public static final int NUM_FINISHED_CHECKPOINTS = 5;
+    private static final int NUM_VERTICES = 3;
+    private static final int PARALLELISM = 4;
+    private static final long CHECKPOINT_INTERVAL_MS = 10;
+
+    public static void main(String[] args) throws RunnerException {
+        Options options = new OptionsBuilder()
+            .verbosity(VerboseMode.NORMAL)
+            .include(UnalignedCheckpointTimeBenchmark.class.getCanonicalName())
+            .build();
+
+        new Runner(options).run();
+    }
+
+    @Benchmark
+    public void unalignedCheckpoint(UnalignedCheckpointEnvironmentContext context) throws Exception {
+        StreamExecutionEnvironment env = context.env;
+        DataStreamSource<byte[]> source = env.addSource(new FiniteCheckpointSource(NUM_FINISHED_CHECKPOINTS));
+        source
+            .slotSharingGroup("source").rebalance()
+            .map((MapFunction<byte[], byte[]>) value -> value).slotSharingGroup("map").rebalance()
+            .addSink(new SlowDiscardSink<>()).slotSharingGroup("sink");
+
+        env.execute();
+    }
+
+    public static class UnalignedCheckpointEnvironmentContext extends FlinkEnvironmentContext {
+
+        @Param({"REMOTE", "LOCAL"})
+        public String mode = "REMOTE";
+
+        @Setup
+        public void setUp() throws IOException {
+            super.setUp();
+
+            env.setParallelism(parallelism);
+            env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Benchmark
+    public void unalignedCheckpointWithRemoteChannel(UCRemoteEnvironmentContext context) throws Exception {
+        unalignedCheckpoint(context);
+    }
+
+    @Benchmark
+    public void unalignedCheckpointWithLocalChannel(UCLocalEnvironmentContext context) throws Exception {
+        unalignedCheckpoint(context);
+    }
```
pnowojski, line 66: Just for the Web UI presentation, what about dropping those methods and introducing a parameter `mode = {"LOCAL", "REMOTE"}` that would configure those two benchmarks? As it is now, this would be presented in the UI as `unalignedCheckpointWithRemoteChannel` and `unalignedCheckpointWithLocalChannel` which is a bit lengthy compared to `unalignedCheckpoint.LOCAL` and `unalignedCheckpoint.REMOTE`.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
```diff
@@ -0,0 +1,164 @@
...
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.io.IOException;
+
+@OperationsPerInvocation(value = UnalignedCheckpointBenchmark.RECORDS_PER_INVOCATION)
```
pnowojski, line 42: This is not correct, you are not using `RECORDS_PER_INVOCATION` to control the number of records per invocation.

pnowojski (commented):
I think we need to take a step back and figure out what do we want to benchmark.

Currently you are benchmarking how long does it take to complete 5 checkpoints, with checkpointing interval of 100ms under the backpressure and to cancel the job. I'm not sure how helpful is that, as if we ever reach checkpointing time below < 100ms, the benchmark results will be capped by the checkpointing interval. From the results that you posted, it looks like checkpointing time is around 5 seconds on your local machine (I guess because of the `SourceFunction` problem), so I guess we are fine for now, but maybe it would be also better to use `FLIP-27` sources here?

It would make more sense for me to check for two different things:
1. how fast quickly can we trigger and complete `N` checkpoints, by setting checkpointing interval to `0ms` (or as small number as possible), make sure the number of concurrent checkpoints is set to `1`, and just measure how quick we can complete those checkpoints (note, we should also in that case modify  `@OperationsPerInvocation` to `N`. This with an induced backpressure.
2. what is the throughput of the job with unaligned checkpoints compared to aligned checkpoints. Set checkpointing interval to some reasonable value and measure how fast can we process `N` records without any backpressure.

What do you think?

zhijiangW:
Thanks for the review and discussions, @pnowojski !

> but maybe it would be also better to use FLIP-27 sources here?

Considering the current situation of mostly used legacy source, it is also meaningful to measure the performance in this case. But I agree we can also supplement FLIP-27 source as an follow-up option.
 
> 1. how fast quickly can we trigger and complete N checkpoints, by setting checkpointing interval to 0ms (or as small number as possible), make sure the number of concurrent checkpoints is set to 1, and just measure how quick we can complete those checkpoints (note, we should also in that case modify @OperationsPerInvocation to N. This with an induced backpressure.

I think my current way is exactly for this purpose, just set `N =5` and the default concurrent cp should be 1 if I remembered correctly. I am not quite sure whether it is really necessary to set even smaller interval since 100ms is already small enough in most practical cases. If we give much smaller values, the backpressure might not be triggered before N checkpoints finished.

> 2. what is the throughput of the job with unaligned checkpoints compared to aligned checkpoints. Set checkpointing interval to some reasonable value and measure how fast can we process N records without any backpressure.

I agree to supplement this case as well, similar with we did in `RemoteBalanceBenchmark`.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
zhijiangW: I will take your option.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
zhijiangW: Maybe my misuse of `RECORDS_PER_INVOCATION` here. My purpose is only for enlarging the results by a reasonable factor to seem readable and comparable, otherwise the raw results will be small enough.

zhijiangW:
> But the problem with the current setup is that if checkpointing time is small, you will be measuring just the time to complete last checkpoint, instead of 5. And the problem is that even if the current interval makes sense, the checkpointing time can change in the future, as we improve things, which would render this benchmark obsolete.

You are right. I misunderstood the semantic of checkpoint interval before. I wrongly thought that the internal was only for triggering the first checkpoint, and the following checkpoint will be triggered immediately after the preceding one completes. As you said, internal `0` should be proper for voiding delays for intermediate checkpoints.

zhijiangW:
> Do you think this can happen? Sources will produce huge number of records immediately (millions of records per second). Even if checkpoint is triggered just a couple of ms after the source started, the task should be already fully back pressured. If this worries you, maybe there is a way of delaying the first checkpoint?

It is also up to how many in-flight buffers we given for network setting. Actually my previous setting of interval 100ms was mainly for delaying the first checkpoint, and ignored the delay of intermediate checkpoints. And I also tested to guarantee that there are input & output inflight buffers spilled during checkpointing.  I would try out other ways to make sure the back pressure happen.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
zhijiangW: Yes, it would be better to adjust `OutputTimeUnit`.

zhijiangW: It seems normal after adjusting to `HOURS`

zhijiangW:
I found that there is a limitation for checkpoint interval which must not be less than 10ms. So it would more make sense if we can guarantee that the checkpointing time is more than 10ms in our benchmark.

pnowojski:
I think `10ms` might be quick enough, that in the foreseeable future checkpoints will not complete that quickly, so IMO it would be good enough to just set `10ms` as the interval. It would also give the those couple of ms to back pressure on the start up.

From this perspective I think maybe it's not worth adding some extra changes in the production code to allow `0ms` checkpoint interval.

zhijiangW:
@pnowojski , I have updated the codes based on the comments!

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
```diff
@@ -48,6 +49,9 @@
 
     private MiniCluster miniCluster;
 
+    @Param({"AlignedCheckpoint", "UnalignedCheckpoint"})
```
pnowojski, line 52: `ALIGNED`/`UNALIGNED` to shorten the benchmark name plus to be consistent with the existing parameters (`ORDERED`/`UNORDERED`, `ROCKS`/`MEMORY`/`FS_ASYNC`/...)?

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
```diff
@@ -80,6 +84,7 @@ public void remoteRebalance(FlinkEnvironmentContext context) throws Exception {
         StreamExecutionEnvironment env = context.env;
         env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
         env.setParallelism(PARALLELISM);
+        env.getCheckpointConfig().enableUnalignedCheckpoints(!mode.equals("AlignedCheckpoint"));
```
pnowojski, line 87: Good idea for re-using this benchmark :)

Before merging this, in order to preserve the benchmark continuity with the previous results, it would be a good idea to rename the `remoteRebalance` benchmark in the codespeed UI to `remoteRebalance.ALIGNED`.

Alternatively after merging, we can remove the `remoteRebalance` benchmark from the UI to clean it up.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
```diff
@@ -0,0 +1,152 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.io.IOException;
+
+import static java.util.concurrent.TimeUnit.HOURS;
+
+@OutputTimeUnit(HOURS)
+public class UnalignedCheckpointBenchmark extends BenchmarkBase {
```
pnowojski, line 46: `@OperationsPerInvocation(NUM_FINISHED_CHECKPOINTS)`?

pnowojski, line 46: nit: add a java doc that this benchmark checks for the time to perform a checkpoint? and/or maybe rename to `UnalignedCheckpointTimeBenchmark`?

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointBenchmark.java
```diff
@@ -0,0 +1,152 @@
...
+            .slotSharingGroup("source").rebalance()
+            .map((MapFunction<byte[], byte[]>) value -> value).slotSharingGroup("map").rebalance()
+            .addSink(new SlowDiscardSink<>()).slotSharingGroup("sink");
+
+        env.execute();
+    }
+
+    public static class UnalignedCheckpointEnvironmentContext extends FlinkEnvironmentContext {
+
+        @Param({"RemoteChannel", "LocalChannel"})
```
pnowojski, line 75: `REMOTE`/`LOCAL`?

pnowojski (commented):
Thanks for the update @zhijiangW . I've left a couple of smaller comments.

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
zhijiangW: Actually I tried the shorten name in early version, but i am wondering it might confuse external users, so i changed to the semantic name finally. Anyway, i have no preference here and will change it as you suggested.

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
zhijiangW: Either option sounds fine to me. But where is the codespeed UI and can I modify it?

src/main/java/org/apache/flink/benchmark/RemoteChannelThroughputBenchmark.java
pnowojski: There are no external users that I know of :)

zhijiangW:
@pnowojski , I have adjusted the `remoteRebalance` to `remoteReblance.ALIGNED` in UI and all the comments are also addressed in separate commit.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
```diff
@@ -42,17 +43,22 @@
 
 import static java.util.concurrent.TimeUnit.HOURS;
 
+/**
+ * The benchmark for measuring the time taken to finish the configured number of
+ * unaligned checkpoints.
+ */
 @OutputTimeUnit(HOURS)
```
pnowojski, line 50: NIT: maybe per minute?

pnowojski (approved):
LGTM % one optional nit

xintongsong:
@pnowojski 
I assume you've assigned me as a reviewer by mis-operation? I'm completely not familiar with the background of this effort.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
zhijiangW: Make sense, already adjusted to `MINUTE`.

zhijiangW:
@xintongsong , I guess it is a mis-operation, ignore. :)

pnowojski:
Yes, sorry @xintongsong !

PR #3, opened 2020-08-26
[FLINK-18905] Provide basic benchmarks for MultipleInputStreamOperator

pnowojski:
```
Benchmark                                         Mode  Cnt      Score      Error   Units
MultipleInputBenchmark.multiInputMapSink         thrpt   15  13311.909 ± 1524.342  ops/ms
MultipleInputBenchmark.multiInputOneIdleMapSink  thrpt   15   7617.621 ±  262.473  ops/ms
```
Results are similar to two input benchmarks, with similar problem of quite large noise between iterations.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
@@ -18,13 +18,15 @@
 
 package org.apache.flink.benchmark;
 
-import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.benchmark.functions.LongSource;
 import org.apache.flink.benchmark.functions.QueuingLongSource;
-import org.apache.flink.benchmark.operators.MultiplyByTwoCoStreamMap;
+import org.apache.flink.benchmark.operators.MultiplyByTwoOperatorFactory;
 import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.datastream.MultipleConnectedStreams;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+import org.apache.flink.streaming.api.transformations.MultipleInputTransformation;
 
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.OperationsPerInvocation;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+		StreamExecutionEnvironment env = context.env;
+
+		env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
+		env.setParallelism(1);
+		env.setRestartStrategy(RestartStrategies.noRestart());
+
+		// Setting buffer timeout to 1 is an attempt to improve twoInputMapSink benchmark stability.
+		// Without 1ms buffer timeout, some JVM forks are much slower then others, making results
+		// unstable and unreliable.
+		env.setBufferTimeout(1);
```
zhijiangW, line 79: I am not quite understand the explanation for this setting, although I found it already existed in previous `TwoInputBenchmark`. And why we do not need it in `multiInputOneIdleMapSink`?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+
+	@Benchmark
+	@OperationsPerInvocation(RECORDS_PER_INVOCATION)
+	public void multiInputMapSink(FlinkEnvironmentContext context) throws Exception {
+
+		StreamExecutionEnvironment env = context.env;
+
+		env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
+		env.setParallelism(1);
+		env.setRestartStrategy(RestartStrategies.noRestart());
```
zhijiangW, line 74: I noticed that `env.setRestartStrategy(RestartStrategies.noRestart());` only exists in this benchmark. Do we also need it in other places besides `TwoInputBenchmark`?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class MultipleInputBenchmark extends BenchmarkBase {
+
+	public static final int RECORDS_PER_INVOCATION = TwoInputBenchmark.RECORDS_PER_INVOCATION;
```
zhijiangW, line 52: Another option is to make `MultipleInputBenchmark` extend `TwoInputBenchmark` and to make use of these fields directly, even we can extract environment setting method to reuse in four specific benchmarks.

```
env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
env.setParallelism(1);
```

It is up to you whether it is worth doing this refactoring.

zhijiangW (commented):
Thanks for the PR @pnowojski . I left some tiny comments. Besides that, I am not sure why we remove the latency benchmark. Could you give some descriptions for the reason in commit `[hotfix] Remove not very useful latency benchmark`?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+		env.addOperator(transform);
+		new MultipleConnectedStreams(env)
+				.transform(transform)
+				.addSink(new DiscardingSink<>());
+	}
+
+	public static class MultiplyByTwoOperatorFactory extends AbstractStreamOperatorFactory<Long> {
+		@Override
+		public <T extends StreamOperator<Long>> T createStreamOperator(StreamOperatorParameters<Long> parameters) {
+			return (T) new MultiplyByTwoOperator(parameters);
```
rkhachatryan, line 127: Can you suppress compiler warning (or fix typing) here and in some other places in this file?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+		private class MultiplyByTwoInput extends AbstractInput<Long, Long> {
+			public MultiplyByTwoInput(
+					AbstractStreamOperatorV2<Long> owner,
+					int inputId) {
+				super(owner, inputId);
+			}
+
+			@Override
+			public void processElement(StreamRecord<Long> element) {
+				output.collect(element.replace(element.getValue() * 2));
```
rkhachatryan, line 157: Should object reuse be disabled for that?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+			return (T) new MultiplyByTwoOperator(parameters);
+		}
+
+		@Override
+		public Class<? extends StreamOperator> getStreamOperatorClass(ClassLoader classLoader) {
+			return MultiplyByTwoOperator.class;
+		}
+	}
+
+	public static class MultiplyByTwoOperator extends AbstractStreamOperatorV2<Long> implements MultipleInputStreamOperator<Long> {
```
rkhachatryan, line 136: I think this class and its factory can be private.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -0,0 +1,161 @@
...
+	@OperationsPerInvocation(ONE_IDLE_RECORDS_PER_INVOCATION)
+	public void multiInputOneIdleMapSink(FlinkEnvironmentContext context) throws Exception {
+
+		StreamExecutionEnvironment env = context.env;
+		env.enableCheckpointing(CHECKPOINT_INTERVAL_MS);
+		env.setParallelism(1);
+
+		QueuingLongSource.reset();
+		DataStreamSource<Long> source1 = env.addSource(new QueuingLongSource(1, ONE_IDLE_RECORDS_PER_INVOCATION - 1));
+		DataStreamSource<Long> source2 = env.addSource(new QueuingLongSource(2, 1));
```
rkhachatryan, line 99: Looking at `QueingLongSource` code, I see it can have only one active source at a time. Right?
I guess that was reasonable for bounded inputs.

But probably it makes sense to benchmark new code with multiple active sources at a time.
WDYT?

rkhachatryan (changes requested):
```
MultipleInputBenchmark.multiInputMapSink         thrpt   15  13311.909 ± 1524.342  ops/ms
```

Why is the error so high - twice as high as 2input? Is it because cnt=15 and local environment?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
rkhachatryan: I also disabled restarts in `ContinuousFileReaderOperatorBenchmark` as it caused it to run indefinitely.
I think it would be helpful to disable it globally, but it's probably out of scope of this PR.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: let's just drop this, I think it hasn't improved anything in the two input benchmark

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: Exactly as @rkhachatryan wrote, it's to prevent live/deadlocking in the benchmarks. I've moved this to the upper level.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: I've removed `setParallelism(1)` as I think that was the default. I'm not sure if it's worth deduplicating this.

pnowojski:
> Why is the error so high - twice as high as 2input? Is it because cnt=15 and local environment?

I don't know, maybe. I think it would be best to verify that over a long term trend.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: My intention is to allow other benchmarks to use this class if needed, in other words, it's not necessarily designed as being private. If someone wants to use it as it is, great. If someone will want to first extract it and move it up, also great.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: That's why we have both `multiInputMapSink` and `multiInputOneIdleMapSink`. The former is using `LongSource`, the later is testing for the case when one input is constantly idle (whitebox benchmarking - we know there is a sensitive code path triggered in that case).

pnowojski:
> Could you give some descriptions for the reason in commit [hotfix] Remove not very useful latency benchmark?

Bah, ok, I will keep it ;) I tried to come up with an explanation but maybe it's better to just keep it as it is. I think this benchmark was causing some troubles when I was trying to move output flusher to mailbox and if I'm dropping some PR I tend to automatically carry on hotfix commits to the next PR.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
rkhachatryan: I see. I would extract it to a separate file then (nit).

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
rkhachatryan: Oh yes, I see now 🤓

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
zhijiangW: That is already fine for me.

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
pnowojski: Object reuse is enabled by default in the benchmarks (`FlinkEnvironmentContext`).

PR #4, opened 2020-10-21
[FLINK-19745][network] Supplement micro-benchmark for bounded blocking partition in remote channel case

zhijiangW:


pnowojski (commented):
Generally speaking LGTM % one rename comment.

One question. Doest your new benchmark has any local channels as well?

However I have one larger question, which set of benchmarks would you like to maintain? After this PR as I understand it, we would have:
- `uncompressedFilePartition` (local)
- `compressedFilePartition` (local)
- `uncompressedMmapPartition` (local)
- `remoteFilePartition`

In this setup, we would be testing the specific variants of the batch partitioning with local setup, and we would have one additional to test for remote.

Alternatively we could:
0. the current status
1. test all variants with remote, additionally to the local
2. have one set of just 3 benchmarks, that are setting up a small job, which has both, a couple of local channels and a handful of remote (just as the real world job would have)
3. both (1. or 0.) AND 2.

I think I'm fine with either option that you would think has the most sense, but let's just make it a conscious decision.

zhijiangW:
@pnowojski , thanks for the quick review and the discussions. I will share some previous thoughts about the above concerns.

> Does your new benchmark has any local channels as well?

The new benchmark is purely remote channels without local, as I setup only one slot per TM.  There are two considerations here:

- Actually we do not really need local channels in my followup network improvements. And the local channels have already been covered in previous `BlockingPartitionBenchmark`.

- I guess it is hard to control the determined behavior if we allow both remote and local channels. E.g. when we run the benchmark multiple times, the number of local channels might be different for every time, which might effect the benchmark results. So I think it would be better and stable to benchmark separately for purely local/remote channels as now.

As for the above alternatives, 

- Option 2: I really considered whether to refactor/make use of previous `BlockingPartitionBenchmark` when introduce this new one. But I found that `BlockingPartitionBenchmark` is mainly for the compression variable for different types(file/mmap), so the remote channel scenario is not necessary for the compression, then I keep the compression variable as a separate one. 

- Option 3: As I explained above I am wondering the mixed remote and local channel might not be stable for every schedule/deployment, then it would affect the benchmark results.

In summary I a bit prefer to the first option if you also agree, then we can see how to adjust them if really necessary future. :)

pnowojski (approved):
Ok, let's leave it as it is :) I see your point.

PR #5, opened 2020-10-29
[FLINK-19884] Add benchmarks for batch-style execution for bounded keyed streams

dawidwys:


src/main/java/org/apache/flink/benchmark/SortingBoundedInputBenchmarks.java
```diff
@@ -0,0 +1,418 @@
...
+		public void processElement(StreamRecord<Tuple2<Integer, byte[]>> element) throws Exception {
+			this.seenRecords++;
+			Integer incomingKey = element.getValue().f0;
+			if (!Objects.equals(incomingKey, currentKey)) {
+				if (!seenKeys.add(incomingKey)) {
+					Assert.fail("Received an out of order key: " + incomingKey);
+				}
+				this.currentKey = incomingKey;
+			}
+		}
```
pnowojski, line 206: nit: maybe extract this code to a helper class `ProcessedKeysOrderAsserter` and re-use it for 1, 2 and 3 input operators?

src/main/java/org/apache/flink/benchmark/SortingBoundedInputBenchmarks.java
```diff
@@ -0,0 +1,418 @@
...
+			Configuration configuration = super.createConfiguration();
+			configuration.set(ExecutionOptions.USE_BATCH_STATE_BACKEND, useSingleKeyStateBackend);
+			configuration.set(ExecutionOptions.RUNTIME_MODE, RuntimeExecutionMode.BATCH);
+			return configuration;
+		}
+	}
+
+	@Benchmark
+	@OperationsPerInvocation(value = RECORDS_PER_INVOCATION)
+	public void oneInputOperator(SortingInputContext context) {
```
pnowojski, line 106: https://github.com/apache/flink-benchmarks/#naming-convention

rename to `sortedOneInput`, `sortedTwoInput`, `sortedMultiInput`?

pnowojski (changes requested):
Can you post the benchmark results from your local machine? The JMH's output with min,max,avg, confidence interval, for example:
```

Result "org.apache.flink.benchmark.KeyByBenchmarks.tupleKeyBy":
  5350.702 ±(99.9%) 239.021 ops/ms [Average]
  (min, avg, max) = (4764.296, 5350.702, 6118.365), stdev = 357.755
  CI (99.9%): [5111.682, 5589.723] (assumes normal distribution)
```
? I mean in order to take a look how well behaved is this benchmark.

Secondly, it would be nice, if you attached some code profiler during the benchmarking procedure, and took a look if you are indeed benchmarking the code that you want to benchmark (for example currently I'm afraid you are benchmarking random number generator and byte's memcopy)

pnowojski:
I've tried to profile the benchmark and:

1. results are very unstable :( I don't know why, maybe because of GC? because...
2. `GC overhead limit exceeded`
```
Iteration  14: 126017.810 ops/ms
Iteration  15: 91305.667 ops/ms
Iteration  16: 43366.582 ops/ms
Iteration  17: <failure>

java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.Arrays.copyOf(Arrays.java:3236)
	at java.io.ByteArrayOutputStream.toByteArray(ByteArrayOutputStream.java:191)
```
It looks like there is some resource leak

src/main/java/org/apache/flink/benchmark/SortingBoundedInputBenchmarks.java
```diff
@@ -0,0 +1,418 @@
...
+				.include(".*" + SortingBoundedInputBenchmarks.class.getCanonicalName() + ".*")
+				.build();
+
+		new Runner(options).run();
+	}
+
+	@State(Thread)
+	public static class SortingInputContext extends FlinkEnvironmentContext {
+		@Param({"true", "false"})
+		public boolean useSingleKeyStateBackend;
```
pnowojski, line 93: Hmmm, do we need to compare it against non single key state backends (`useSingleKeyStateBackend = false`)? (We already have benchmarks for one/two/multiple input operators) ?

src/main/java/org/apache/flink/benchmark/SortingBoundedInputBenchmarks.java
```diff
@@ -0,0 +1,418 @@
...
+
+		SingleOutputStreamOperator<Long> counts = elements
+			.keyBy(element -> element.f0)
+			.transform(
+				"Asserting operator",
+				BasicTypeInfo.LONG_TYPE_INFO,
+				new AssertingOperator()
+			);
+
+		counts.addSink(new DiscardingSink<>());
```
pnowojski, line 122: aren't we missing `env.execute()`? The benchmark results I observed are suspiciously high
```
Iteration   1: 125200.947 ops/ms
Iteration   2: 164695.613 ops/ms
Iteration   3: 133524.562 ops/ms
```
This can also explain resource leaks and stability problems.

pnowojski (changes requested):

dawidwys:
First of all sorry for a poor quality of the initial contribution. It has way too many problems.

I tried addressing all your comments. Here are the results from an execution on my local machine:

```
Result "org.apache.flink.benchmark.SortingBoundedInputBenchmarks.sortedMultiInput":
  1397.457 ±(99.9%) 41.128 ops/ms [Average]
  (min, avg, max) = (1217.022, 1397.457, 1485.721), stdev = 61.558
  CI (99.9%): [1356.330, 1438.585] (assumes normal distribution)

Result "org.apache.flink.benchmark.SortingBoundedInputBenchmarks.sortedOneInput":
  1821.521 ±(99.9%) 29.967 ops/ms [Average]
  (min, avg, max) = (1717.496, 1821.521, 1903.986), stdev = 44.853
  CI (99.9%): [1791.554, 1851.488] (assumes normal distribution)

Result "org.apache.flink.benchmark.SortingBoundedInputBenchmarks.sortedTwoInput":
  1444.770 ±(99.9%) 35.563 ops/ms [Average]
  (min, avg, max) = (1348.234, 1444.770, 1530.831), stdev = 53.230
  CI (99.9%): [1409.206, 1480.333] (assumes normal distribution)
```

src/main/java/org/apache/flink/benchmark/SortingBoundedInputBenchmarks.java
dawidwys: The difference is in a state backend that is used. In both cases I do sort the inputs, so it is different from plain benchmarks for one/two/multiple operators. The difference between the "special" state backend vs memory state backend is not big enough to manifest in the tests. Therefore I removed the flag as per your suggestion.

pnowojski (approved):
Thanks for the updates LGTM

PR #6, opened 2020-11-03
[FLINK-19680] Explicitly set alignment timeout to 0 in the existing Unaligned Checkpoints benchmarks

pnowojski:
This should fix the current benchmark time outing and will make the benchmark independent of any potential future default value changes.


rkhachatryan (approved):
LGTM

PR #8, opened 2021-03-03
[FLINK-21526][runtime] Replace scheduler benchmarks with wrapper classes

Thesharing:
Due to [FLINK-21514](https://issues.apache.org/jira/browse/FLINK-21514), we find that when someone modifies interfaces and methods used by scheduler benchmarks, the compilation and execution of flink-benchmark may break. 

To improve the stability of scheduler benchmarks, we decide to replace scheduler benchmarks with wrapper/executor classes, and move the current implementations to the flink repository. Also we'll implement several unit tests based on the implementations. In this way, when someone modifies interfaces and methods used by scheduler benchmarks, the unit tests fails, and then they must fix the broken benchmarks. 

After moving the implementations of scheduler benchmarks to the Flink repository, we replace scheduler benchmarks in flink-benchmark with wrapper classes.

For more details about this pull request, please check [FLINK-21524](https://issues.apache.org/jira/browse/FLINK-21524) and [FLINK-21526](https://issues.apache.org/jira/browse/FLINK-21526).

For more details about the scheduler benchmarks, please check [FLINK-20612](https://issues.apache.org/jira/browse/FLINK-20612).

PR #10, opened 2021-03-08
[FLINK-21658] Adopts JobGraphBuilder to create job graph

Myasuka:
This PR avoid current broken compilation.


zhuzhurk:
SchedulerBenchmarkUtils has been migrated to flink repo in https://github.com/apache/flink/pull/15075 and will soon be removed from flink-benchmarks in #8.
Therefore I think we do need this fix.

zhuzhurk:
Now I changed my mind. #8 requires flink artifacts to be deployed with changes from apache/flink#15075. It may still need some time. To unblock flink benchmarks, let's merge this fix first.

zhuzhurk (approved):

PR #13, opened 2021-06-08
[FLINK-22881] Parameterize InputBenchmark test to use FLIP-27 source

dawidwys:


src/main/java/org/apache/flink/benchmark/InputBenchmark.java
```diff
@@ -34,31 +35,33 @@
...
-			.include(".*" + InputBenchmark.class.getCanonicalName() + ".*")
-			.build();
+				.verbosity(VerboseMode.NORMAL)
+				.include(".*" + InputBenchmark.class.getCanonicalName() + ".*")
+				.build();
 
 		new Runner(options).run();
 	}
 
+	@Param({"LEGACY_SOURCE", "FLIP_27_SOURCE_BOUNDED", "FLIP_27_SOURCE_UNBOUNDED"})
```
pnowojski, line 51: This will become a part of the benchmark name, so I would shorten it as much as possible.

```
{"LEGACY", "BOUNDED", "UNBOUNDED"}
```
```
{"LEGACY", "F27_BOUNDED", "F27_UNBOUNDED"}
```
?

src/main/java/org/apache/flink/benchmark/functions/LongSourceType.java
```diff
@@ -0,0 +1,56 @@
...
+	}),
+	FLIP_27_SOURCE_BOUNDED((env, maxValue) -> {
+		return env.fromSource(
+				new LongNewSource(Boundedness.BOUNDED, maxValue),
+				WatermarkStrategy.noWatermarks(),
+				"NewLongSource");
+	}),
+	FLIP_27_SOURCE_UNBOUNDED((env, maxValue) -> {
+		return env.fromSource(
+				new LongNewSource(Boundedness.BOUNDED, maxValue),
```
pnowojski, line 43: `FLIP_27_SOURCE_UNBOUNDED` & `BOUNDED`? How does it differ compared to `FLIP_27_SOURCE_BOUNDED`? Do we need both? Is there a runtime difference between bounded & unbounded?

src/main/java/org/apache/flink/benchmark/functions/LongNewSource.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * The source should produce same records as {@link LongSource}.
+ */
+public class LongNewSource extends NumberSequenceSource {
+	private final Boundedness boundedness;
+	private final long maxValue;
+
+	public LongNewSource(Boundedness boundedness, long maxValue) {
+		super(-1, -1); // we do not use the from/to of the underlying source
```
pnowojski, line 40: What's wrong with `NumberSequenceSource`? Can not we use it as it is?

pnowojski (commented):
Thanks for providing those benchmarks. I've left a couple of questions.

pnowojski:
Have you seen CI failure? It looks unrelated, but if this will happen again, it's an issue that we will need to fix 🙄

src/main/java/org/apache/flink/benchmark/functions/LongSourceType.java
dawidwys: Yes, for `BOUNDED` we might use the sorting `DataOutputs/Inputs`, use the pipelined scheduling, use different operators etc.

src/main/java/org/apache/flink/benchmark/functions/LongNewSource.java
dawidwys: I wanted to have exact same logic as the `LongSource`. `LongSource` generates records from 0 to maxValue for every parallel instance.

The `NumberSequenceSource` would split the range 0 to `maxValue` between all subtasks.

dawidwys:
Yes, I've looked into it. It failed when downloading from maven.

src/main/java/org/apache/flink/benchmark/functions/LongNewSource.java
dawidwys: As a side comment. I tried reusing only the reader and enumerator, but the problem is that serializers for splits and enumerator checkpoints are private.

src/main/java/org/apache/flink/benchmark/functions/LongSourceType.java
pnowojski: By first question I meant that there is a bug with both BOUNDED and UNBOUNDED enum setting `Boundedness.BOUNDED` parameter.

Don't we have other benchmarks that are covering the case of sorting outputs/inputs?

src/main/java/org/apache/flink/benchmark/functions/LongNewSource.java
pnowojski: 👍

pnowojski:
One last question if you are 100% sure that we need the sorting benchmarks (keep in mind that we are already bottlenecking the benchmarking pipeline a bit). If the answer is yes, LGTM.

src/main/java/org/apache/flink/benchmark/functions/LongSourceType.java
dawidwys: We do. I'd still have the enum here. I am fine though with dropping the `F27_BOUNDED` from `InputBenchmark`.

I spotted the bug after your comment and fixed it.

PR #14, opened 2021-06-11
[FLINK-22973] Provide unaligned checkpoint with timeouts benchmark

pnowojski:


src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
```diff
@@ -42,119 +43,90 @@
...
             Configuration conf = super.createConfiguration();
-
-            if (mode.equals("REMOTE")) {
-                conf.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, 1);
-                conf.setInteger(ConfigConstants.LOCAL_NUMBER_TASK_MANAGER, NUM_VERTICES * PARALLELISM);
-            } else {
-                conf.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, NUM_VERTICES * PARALLELISM);
-                conf.setInteger(ConfigConstants.LOCAL_NUMBER_TASK_MANAGER, 1);
-            }
+            conf.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, 1);
```
akalash, line 114: Don't we want to test different numbers of slots per task manager? Or it doesn't make a lot of sense in this scenario?

akalash (approved):
I left a couple of minor questions there. But anyway LGTM.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
```diff
@@ -42,119 +43,90 @@
...
+                .slotSharingGroup("sink");
 
         env.execute();
     }
 
     public static class UnalignedCheckpointEnvironmentContext extends FlinkEnvironmentContext {
 
-        @Param({"REMOTE", "LOCAL"})
-        public String mode = "REMOTE";
+        @Param({"0", "1", "ALIGNED"})
```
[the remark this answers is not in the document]
pnowojski: I think so. `ALIGNED` and `0` do not test timeouts whatsoever, so we need some setting that would benchmark the timeouts. The lowest possible timeout setting (`1ms`) makes the most sense, as we would ideally be expecting the results be almost as good (+/- 1ms) as fully unaligned checkpoints. Any setting higher than that, and we are actually benchmarking sleep times, instead of efficiency/overhead of the time outing algorithm.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
pnowojski: With infinite resources available it would make sense to test with various number of local/remote input channels. But having as many as possible remote input channels and single local channel I think is a worst case scenario that's testing both code paths at the same time, so single slot per TM I think should be good enough, and should cover all cases well enough.

If needed, we can expand this test in the future.

PR #16, opened 2021-06-22
[hotfix] Remove SchedulerBenchmark#teardown

zentol:
It should be unnecessary to shut down the executor because as far as I can tell it isn't used.


pnowojski (approved):
LGTM % green travis

PR #17, opened 2021-06-25
[FLINK-23078] Fix SchedulerBenchmarkExecutorBase#teardown

Thesharing:
In FLINK-22988, TestUtils are ported from Scala to Java. However, `SchedulerBenchmarkBase` in flink-benchmark calls this method when each benchmark teardowns. This make SchedulerBenchmarks not compiling, as FLINK-23078 says. To solve this issue, we'd like to wrap `TestingUtils.defaultExecutor().shutdownNow()` into `SchedulerBenchmarkUtil`. If this method is changed in the future, its reference must be fixed, too. 

Furthermore, it's confusing to have `SchedulerBenchmarkBase` in both flink and flink-benchmark. So I'd like to propose change `SchedulerBenchmarkBase` into `SchedulerBenchmarkExecutorBase` in flink-benchmark.

### Brief changelog

* Change `TestingUtils.defaultExecutor().shutdownNow()` into `SchedulerBenchmarkUtil#shutdownTestingUtilDefaultExecutor`.

### DIFF
--- a/src/main/java/org/apache/flink/scheduler/benchmark/SchedulerBenchmarkExecutorBase.java
+++ b/src/main/java/org/apache/flink/scheduler/benchmark/SchedulerBenchmarkExecutorBase.java
@@ -45,7 +45,7 @@ import java.util.concurrent.TimeUnit;
 		"-Dcom.sun.management.jmxremote.ssl=false",
 		"-Dcom.sun.management.jmxremote.ssl"
 })
-public class SchedulerBenchmarkBase {
+public class SchedulerBenchmarkExecutorBase {
 
 	public static void runBenchmark(Class<?> clazz) throws RunnerException {
 		Options options = new OptionsBuilder()
--- a/src/main/java/org/apache/flink/scheduler/benchmark/deploying/DeployingDownstreamTasksInBatchJobBenchmarkExecutor.java
+++ b/src/main/java/org/apache/flink/scheduler/benchmark/deploying/DeployingDownstreamTasksInBatchJobBenchmarkExecutor.java
@@ -21,7 +21,7 @@ package org.apache.flink.scheduler.benchmark.deploying;
 import org.apache.flink.runtime.executiongraph.Execution;
 import org.apache.flink.runtime.scheduler.benchmark.JobConfiguration;
 import org.apache.flink.runtime.scheduler.benchmark.deploying.DeployingDownstreamTasksInBatchJobBenchmark;
-import org.apache.flink.scheduler.benchmark.SchedulerBenchmarkBase;
+import org.apache.flink.scheduler.benchmark.SchedulerBenchmarkExecutorBase;
 
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
@@ -35,7 +35,7 @@ import org.openjdk.jmh.runner.RunnerException;
  * The benchmark of deploying downstream tasks in a BATCH job.
  * The related method is {@link Execution#deploy}.
  */
-public class DeployingDownstreamTasksInBatchJobBenchmarkExecutor extends SchedulerBenchmarkBase {
+public class DeployingDownstreamTasksInBatchJobBenchmarkExecutor extends SchedulerBenchmarkExecutorBase {
 
 	@Param("BATCH")
 	private JobConfiguration jobConfiguration;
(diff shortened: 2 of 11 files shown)


src/main/java/org/apache/flink/scheduler/benchmark/SchedulerBenchmarkExecutorBase.java
```diff
@@ -48,7 +48,7 @@
 		"-Dcom.sun.management.jmxremote.ssl=false",
 		"-Dcom.sun.management.jmxremote.ssl"
 })
-public class SchedulerBenchmarkBase {
+public class SchedulerBenchmarkExecutorBase {
```
pnowojski, line 51: +1 for renaming `SchedulerBenchmarkExecutorBase`

src/main/java/org/apache/flink/scheduler/benchmark/SchedulerBenchmarkBase.java
```diff
@@ -55,4 +58,10 @@ public static void runBenchmark(Class<?> clazz) throws RunnerException {
 
 		new Runner(options).run();
 	}
+
+
+	@TearDown
+	public void teardown() {
+		SchedulerBenchmarkUtils.shutdownTestingUtilDefaultExecutor();
```
pnowojski, line 65: Sorry, I'm confused. Why do we need this? It's again the same problem as with old `TestingUtils`. It's a class defined in `flink` repository, which is not part of the `Flink's` public API. Why can not we call this `shutdownTestingUtilDefaultExecutor()` in `org.apache.flink.runtime.scheduler.benchmark.e2e.SchedulerBenchmarkBase#teardown`?

Thesharing: Sorry for making it confusing. 
1. `TestingUtils.defaultExecutor()` is used in `DefaultSchedulerBuilder` by default. Since we use this builder to build a scheduler directly, we need to call `TestingUtils.defaultExecutor().shutdownNow()` to avoid the leak of the executor.
2. SchedulerBenchmarkBase is only inherited by benchmarks located in `org.apache.flink.runtime.scheduler.benchmark.e2e`. To make it simple, I wrap `TestingUtils.defaultExecutor().shutdownNow();` into `SchedulerBenchmarkUtils` and call this method in `SchedulerBencmarkExecutorBase`. In this way we can make sure when `TestingUtils` is modified, this reference will be fixed together.

zhuzhurk: I agree with @pnowojski 's suggestion.
Maybe we can let all the scheduler benchmarks explicitly set the `executor` when creating a Scheduler/ExecutionGraph. To be specific, the benchmarks can extend a benchmark base which host an `executor` for Scheduler/ExecutionGraph creation and provides a `shutDown()` method.  All the benchmark executors in `flink-benchmarks` should invoke the `shutDown()` when tearing down.
This also allows the benchmarks to get rid of the implicit assumption that they relies on `TestingUtils.defaultExecutor()`, which may change in the future and break this benchmarks.

pnowojski: Thanks for the explanation @Thesharing  and +1 for what @zhuzhurk wrote.

Generally speaking, based on our previous experience, if benchmark is not using public API it should be fully implemented in the `flink` repository, while keeping `flink-benchmarks` as simple as possible for as much compatibility as possible. For example having exposed across those two repositories only very stable three methods: `#execute()`, `#setup()` (this one optionally as builder if there are parameters), `#tearDown()`.

Thesharing: Thank you for your suggestions, @zhuzhurk and @pnowojski. I came up with a pull request: https://github.com/apache/flink/pull/16308. In this pull request all the scheduler benchmarks inherit a base class. And in the base class a `setup` and `teardown` method is added. The default executor is closed in the `teardown` method.

pnowojski: Thanks for the fix. I see it has already been reviewed. As I understand, in that case we can just drop the first commit from this PR and merge the rename?

Thesharing: Aside from the rename, we still need to add a `teardown` method to all scheduler benchmarks.

zhuzhurk (approved):
LGTM.

PR #18, opened 2021-06-25
[FLINK-23153] Fix the reference of FutureUtils in RecordSource

Thesharing:
In FLINK-23085, FutureUtils is moved from flink-runtime to flink-core. The reference in flink-benchmark should also be changed. The reference is located at: org/apache/flink/benchmark/operators/RecordSource.java.


Thesharing:
After offline discussion with @zhuzhurk, we both agree that the reference of FutureUtils here could be removed by replace `FutureUtils.completedVoidFuture()` with `CompletableFuture.completedFuture(null)`.

dawidwys:
LGTM, merging.

PR #19, opened 2021-07-06
[FLINK-23284] Reset the TARGET_COUNT_REACHED_LATCH before each execution of the job

dawidwys:


pnowojski (commented):
What was the issue? That the latch was being reseted once per iteration instead of invocation? Where iteration is one or more invocations? And this latch was effectively working only for the first invocation? If so:

1. What about keeping `tearDown()` method but just using `Level.Invocation`?
2. Add this explanation to the commit message?

pnowojski:
On a second thought. After reading all of the warnings around `Level.Invocation` I'm actually fine with the current approach. `Level.Invocation` is scary.

dawidwys:
Generally speaking I don't know JMH enough to make a good judgement why was it so. I don't know how the `Levels` work and if 1. is enough :(  Why do you prefer using the `tearDown` method? I doubt calling the `reset` inside the benchmarked method introduces significant time penalty.

dawidwys:
BTW, frankly speaking, the `Level.Iteration` should be enough from the javadoc, but apparently it isn't. :(

pnowojski:
Why `Iteration` should be enough? If you have many invocation within one single iteration, this latch would have been reseted only once. Your code is equivalent of `tearDown` with `Invocation` level, but according to this scary java doc, `Level.Invocation` should be avoided - and yes, in our case it doesn't matter if you include the reset in the benchmark invocation.

pnowojski (approved):
LGTM % comment in the commit

dawidwys:
After an offline discussion with Piotr, I realized I do not understand what an `Invocation` and an `Iteration` are. Piotr explained to me that within an `Invocation`, there might be multiple calls to the benchmarked method (`readFileSplit` in this case). If there are multiple invocations of the method within a single iteration, the second invocation did not have the latch properly reset, resulting in all records being emitted in the finishing stage of the `ContinuousFileReadingOperator`. This led to testing a different behaviour than the envisioned.

PR #20, opened 2021-07-14
[FLINK-23383] Prevent duplicate setUp/tearDown calls 

zentol:
Prevents an issue where `FlinkEnvironmentContext#setUp/tearDown` would be called multiple times, because subclasses contained overrides annotated with `@SetUp/@TearDown` which called into the parent version.

The effect can be easily seen by adding some `println` statements to `FlinkEnvironmentContext#setUp` and running the `WindowBenchmarks` (or any benchmark with a `FlinkEnvironmentContext` subclass).


pnowojski (approved):
Lol. Nice catch. LGTM assuming green travis.

PR #22, opened 2021-07-15
[FLINK-23392] Exclude startup/shutdown time from benchmarks

zentol:
Sets up a MiniCluster in the `FlinkEnvironmentContext`, which all jobs will be run against.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
@@ -109,11 +109,19 @@ public class UnalignedCheckpointTimeBenchmark extends BenchmarkBase {
             }
         }
 
+        @Override
+        protected int getNumberOfSlotsPerTaskManager() {
+            return 1;
+        }
+
+        @Override
+        protected int getNumberOfTaskManagers() {
+            return NUM_VERTICES * PARALLELISM;
+        }
+
         protected Configuration createConfiguration() {
             Configuration conf = super.createConfiguration();
-            conf.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, 1);
             conf.set(TaskManagerOptions.MEMORY_SEGMENT_SIZE, new MemorySize(1024 * 4));
-            conf.setInteger(ConfigConstants.LOCAL_NUMBER_TASK_MANAGER, NUM_VERTICES * PARALLELISM);
             return conf;
         }
     }
--- a/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
@@ -116,6 +116,7 @@ public class UnalignedCheckpointTimeBenchmark extends BenchmarkBase {
 
         @Override
         protected int getNumberOfTaskManagers() {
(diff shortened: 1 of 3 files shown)


pnowojski (approved):
Change LGTM. Before merging, I would like to schedule a benchmark-request build of this PR and https://github.com/apache/flink-benchmarks/pull/21 to verify that they do not cause any regressions.

However currently benchmark worker machine is scheduled for a downtime as someone else is also fixing another problem with the benchmarks 🙄 ( https://issues.apache.org/jira/browse/FLINK-23398 ), I will need to coordinate the efforts. Anyway, please hold on with merging until this is tested.

zentol:
> Anyway, please hold on with merging until this is tested.

Of course.

Looks like CI is failing because some tests need 4 slots, but the MiniCluster is only started with 1. The LocalExecutor used so far automatically sets the number of slots to the highest parallelism of the job; we now need to make this explicit.

pnowojski:
In that case shouldn't we pass the number of slots from the benchmark in a parameter? Is there any drawback of having configured more slots that are unused?

zentol:
> In that case shouldn't we pass the number of slots from the benchmark in a parameter?

Ideally, yes, but then we would to touch every benchmark, and I don't know the nicest way to pass a parameter into the FlinkEnvironmentContext (aside from subclasses every where).

> Is there any drawback of having configured more slots that are unused?

I don't think so;

pnowojski:
>  and I don't know the nicest way to pass a parameter into the FlinkEnvironmentContext (aside from subclasses every where).

An overloaded `setUp(int taskSlots)` method? But I'm fine with having default 4 task slots for now as long as it doesn't cause problems for benchmarks that so far were using fewer slots.

zentol:
> But I'm fine with having default 4 task slots for now as long as it doesn't cause problems for benchmarks that so far were using fewer slots.

We can test the impact by explicitly configuring `TaskManagerOptions.NUM_TASK_SLOTS` in the FlinkEnvironmentContext, which the LocalExecutor would honor.

zentol:
So we run 1 benchmark with the current master + explicitly setting the number of slots to 4.
If we don't see changes we can be pragmatic and always set the number of slots to 4, if we see changes we can look into modifying all benchmarks to set the number of slots.

> An overloaded setUp(int taskSlots) method?

That would require subclasses in each benchmark, no? Because the benchmarks extend the `BenchmarkBase`, not the `FlinkEnvironmentContext`.

src/main/java/org/apache/flink/benchmark/UnalignedCheckpointTimeBenchmark.java
```diff
@@ -116,6 +116,7 @@ protected int getNumberOfSlotsPerTaskManager() {
 
         @Override
         protected int getNumberOfTaskManagers() {
+            // why is this using PARALLELISM when we don't actually use it?
```
pnowojski, line 119: this is probably a bug, `PARALLELISM` should have been used instead of `parallelism` in `env.setParallelism(parallelism);`. But let's keep it for now for continuity of the benchmark results while we are merging this? I can fix it later on.

pnowojski (approved):
I've scheduled another comparison build
http://codespeed.dak8s.net:8080/job/flink-benchmark-request/307/
but I'm also fine with merging this as it is (after squashing), and we can see the trend of the `UnalginedCheckpointTimeBenchmark`s

pnowojski (approved):
LGTM let's merge this

PR #24, opened 2021-07-15
[FLINK-23398] Exclude signature files from shaded benchmarks.jar

Myasuka:
This fix current check when running benchmarks on the shaded jar package.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -381,7 +381,7 @@ under the License.
 									</transformers>
 									<filters>
 										<filter>
-											<artifact>*:*</artifact>
+											<artifact>*</artifact>
 											<excludes>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -379,6 +379,16 @@ under the License.
 		                                                                        <projectName>Apache Flink</projectName>
 		                                                                </transformer>
 									</transformers>
+									<filters>
+										<filter>
+											<artifact>*:*</artifact>
```
zentol, line 384:
```suggestion
											<artifact>*</artifact>
```
nit: this works as well ;)

zentol (approved):
This looks fine to me.

pnowojski (approved):

PR #25, opened 2021-07-19
[FLINK-23208] Add a benchmark for processing timers

Jiayi-Liao:
This pull request adds a new benchmark for processing timers. 

Output example: 

```
Result "org.apache.flink.benchmark.ProcessingTimerBenchmark.fireProcessingTimers":
  0.683 ±(99.9%) 0.010 ops/ms [Average]
  (min, avg, max) = (0.659, 0.683, 0.730), stdev = 0.015
  CI (99.9%): [0.673, 0.692] (assumes normal distribution)


# Run complete. Total time: 00:01:37

Benchmark                                       Mode  Cnt  Score   Error   Units
ProcessingTimerBenchmark.fireProcessingTimers  thrpt   30  0.683 ± 0.010  ops/ms

Benchmark result is saved to jmh-result.csv
```

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.core.testutils.OneShotLatch;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.streaming.api.functions.KeyedProcessFunction;
+import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
+import org.apache.flink.streaming.api.functions.source.RichParallelSourceFunction;
+import org.apache.flink.util.Collector;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Random;
+
+@OperationsPerInvocation(value = ProcessingTimerBenchmark.PROCESSING_TIMERS_PER_INVOCATION)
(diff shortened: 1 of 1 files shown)


Jiayi-Liao:
@pnowojski  1_000_000 timers are too many for current processing timer's mechanism. The benchmark consumes 4_000 timers with four subtasks. Please take a review when you have time, thanks :)

src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+
+public class ProcessingTimerBenchmark extends BenchmarkBase {
+
+    private static final int PARALLELISM = 4;
```
pnowojski, line 39: `PARALLELISM = 1` and reduce the timers per invocation to `1000`? I think parallelism doesn't add any more coverage in this benchmark and just complicates the setup and increases start up costs?

src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+                .include(".*" + ProcessingTimerBenchmark.class.getCanonicalName() + ".*")
+                .build();
+
+        new Runner(options).run();
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(value = ProcessingTimerBenchmark.PROCESSING_TIMERS_PER_INVOCATION)
+    public void fireProcessingTimers(FlinkEnvironmentContext context) throws Exception {
+        LATCH = new CountDownLatch(PROCESSING_TIMERS_PER_INVOCATION);
```
pnowojski, line 58: smaller issue: with parallelism of 1, I would replace this with single shot latch and a non-volatile counter. With just a single operator firing timers, that single operator wouldn't need to synchronise with anyone to decide whether it's time to trigger the latch or not and it would avoid synchronisation overheads for the count down.

src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+        Options options = new OptionsBuilder()
+                .verbosity(VerboseMode.NORMAL)
+                .include(".*" + ProcessingTimerBenchmark.class.getCanonicalName() + ".*")
+                .build();
+
+        new Runner(options).run();
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(value = ProcessingTimerBenchmark.PROCESSING_TIMERS_PER_INVOCATION)
```
pnowojski, line 56: nitty nit: maybe move this to the top of the class for the sake of consistency with other benchmarks?

src/main/java/org/apache/flink/benchmark/ProcessingTimerBenchmark.java
```diff
@@ -0,0 +1,115 @@
...
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+
+public class ProcessingTimerBenchmark extends BenchmarkBase {
+
+    private static final int PARALLELISM = 4;
+    private static final int PROCESSING_TIMERS_PER_INVOCATION = 4_000;
+    private static final int PROCESSING_TIMERS_PER_RECORD = 4_000;
```
pnowojski, line 41: Do we need both of those constants? Can not we use a single one for both as we have always just a single record?

pnowojski (commented):
Thanks for the benchmark, code looks mostly good to me with a couple of minor comments.

Jiayi-Liao:
@pnowojski Thanks for the review. I've addressed all the comments and updated the description.

pnowojski (approved):
Thanks for the update, LGTM :) can you squash the commits before merging

Jiayi-Liao:
> Thanks for the update, LGTM :) can you squash the commits before merging

Done.

pnowojski:
Thanks for the explanation @zhiyisun . Option 2. sounds to me preferable.

PR #27, opened 2021-07-26
[FLINK-23491] Update flink-runtime dependency

zentol:
Removes the scala suffix from the flink-runtime dependency, which is required because flink-runtime on master is now scala-free.

Because of the suffix the benchmarks were using an old snapshot test-jar, which referenced classes that no longer exist.

Note that CI may fail because Flink snapshots are currently broken.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,8 +46,8 @@ under the License.
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<flink.version>1.14-SNAPSHOT</flink.version>
-		<flink.shaded.version>10.0</flink.shaded.version>
-		<netty.tcnative.version>2.0.25.Final</netty.tcnative.version>
+		<flink.shaded.version>14.0</flink.shaded.version>
+		<netty.tcnative.version>2.0.39.Final</netty.tcnative.version>
 		<java.version>1.8</java.version>
 		<scala.binary.version>2.11</scala.binary.version>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
--- a/pom.xml
+++ b/pom.xml
@@ -104,13 +104,6 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-runtime_${scala.binary.version}</artifactId>
-			<version>${flink.version}</version>
-			<type>test-jar</type>
-		</dependency>
-
-		<dependency>
-			<groupId>org.apache.flink</groupId>
 			<artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
 			<version>${flink.version}</version>
 			<type>test-jar</type>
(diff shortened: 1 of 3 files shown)


zhuzhurk (approved):
+1

pnowojski (approved):
"funny" bug

Does it mean any benchmark build that we executed in the last couple of weeks was flawed and was always benchmarking the old `flink-runtime` code?

pom.xml
```diff
@@ -104,7 +104,7 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-runtime_${scala.binary.version}</artifactId>
+			<artifactId>flink-runtime</artifactId>
 			<version>${flink.version}</version>
 			<type>test-jar</type>
 		</dependency>
```
pnowojski, line 110: I've tried removing this whole dependency in order to locate benchmarks that have been affected by this bug, and the code was still compiling? Am I missing something?

zentol:
Does it mean any benchmark build that we executed in the last couple of weeks was flawed and was always benchmarking the old flink-runtime code? No; if they ran they they used the new flink-runtime.

pom.xml
zentol: it could actually be that we don't need it. If they compile anyway then we should be able to remove it.

zentol: The benchmark base is actually in the test-jar.

zhuzhurk: There are two `flink-streaming-java` dependencies. One of default scope.
May it be that there are two `flink-runtime` dependencies which was introduced by `flink-streaming-java` and this direct `flink-runtime` dependency. And conflicts happened?

zentol: Without this dependency we it as a transitive dependency:
```
[INFO] +- org.apache.flink:flink-test-utils_2.11:jar:1.14-SNAPSHOT:compile
[INFO] |  +- org.apache.flink:flink-runtime:test-jar:tests:1.14-SNAPSHOT:compile
```

zentol: So we had 2 different flink-runtime test-jar dependencies on the CP, but only one of them could fully work because only 1 version of flink-runtime was there.

zentol: TL;DR: let's just remove it for now. (although I'm not a huge fan of used undeclared dependencies, precisely because of such surprises...)

zentol:
urgh, looks like there's a weird issue with Travis that results in TMP_DIRS not being properly configured; looking into it...doesn't seem to affect the jenkins builds.

zentol:
The test failure is related to FLINK-23499. The path `file:///<tmp_dir>` is split into `file` and `<tmp_dir>`.

Removing the scheme fixes the issue, but we should also look into hardening the akka extraction mechanism.

zentol:
And _another_ issue popped up because the netty-native version is outdated and not compatible with the flink-shaded-netty version used by Flink.

zentol:
Previous green CI build (I just amended the commit messages): https://app.travis-ci.com/github/apache/flink-benchmarks/builds/233660784

pnowojski (approved):

PR #29, opened 2021-08-10
[FLINK-23689] Increase ProcessingTimerBenchmark length

pnowojski:
Previous 1000 records was forced upon this benchmark because of
very slow production code. After resolving FLINK-23689 we can
safely increase number of records. Otherwise we would be testing
setup time of a Flink job.


pnowojski:
@Jiayi-Liao could you take a look? 

Increasing beyond 100_000 records stops showing any performance improvement, so the setup costs are already negligible. With 1_000 records benchmark results were around ~7ops/ms. With 150_000 records, results are ~250ops/ms.

Jiayi-Liao:
@pnowojski Good point! 

LGTM

PR #31, opened 2021-09-16
[FLINK-24300] Added benchmarks for idling chained sources

dawidwys:


src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -82,10 +98,85 @@ public void multiInputOneIdleMapSink(FlinkEnvironmentContext context) throws Exc
 		env.execute();
 	}
 
+	@Benchmark
+	@OperationsPerInvocation(CHAINED_IDLE_RECORDS_PER_INVOCATION)
+	public void multiInputIdleSourcesChainedWithInput(FlinkEnvironmentContext context) throws Exception {
```
pnowojski, line 103: too long benchmark name

`multiInputChainedIdleSource`?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -36,18 +50,20 @@
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 import org.openjdk.jmh.runner.options.VerboseMode;
 
-public class MultipleInputBenchmark extends BenchmarkBase {
+import java.util.concurrent.CompletableFuture;
 
+public class MultipleInputBenchmark extends BenchmarkBase {
 	public static final int RECORDS_PER_INVOCATION = TwoInputBenchmark.RECORDS_PER_INVOCATION;
 	public static final int ONE_IDLE_RECORDS_PER_INVOCATION = TwoInputBenchmark.ONE_IDLE_RECORDS_PER_INVOCATION;
+	public static final int CHAINED_IDLE_RECORDS_PER_INVOCATION = 3000;
```
pnowojski, line 58: Why only 3000? That's 5000 times less then the other benchmarks here, so something sound very wrong here.

Also are you sure https://github.com/apache/flink/pull/17303 is fixing the issue? I've tried it locally and couldn't see a difference in this benchmark with/without this change. What are the results that you have meassured?

dawidwys: 1. I forgot to move a flag from the UT and thus the chaining was not applied in the benchmark, sorry for that.
2. Yes, I am sure it fixes it. With the flag you can see a difference
3. I picked a significantly smaller number to give the benchmark a chance to finish with the old code. With 6 digit number of records, I am afraid it would not finish (with 5 digits it already goes into minutes for a single iteration)

dawidwys: 3000 records
* without the fix:
```
Result "org.apache.flink.benchmark.MultipleInputBenchmark.multiInputChainedIdleSource":
  4.352 ±(99.9%) 0.174 ops/ms [Average]
  (min, avg, max) = (4.003, 4.352, 5.241), stdev = 0.260
  CI (99.9%): [4.179, 4.526] (assumes normal distribution)
```
* with the fix
```
Result "org.apache.flink.benchmark.MultipleInputBenchmark.multiInputChainedIdleSource":
  7.802 ±(99.9%) 0.082 ops/ms [Average]
  (min, avg, max) = (7.608, 7.802, 8.243), stdev = 0.123
  CI (99.9%): [7.720, 7.884] (assumes normal distribution)
```

pnowojski: Why is the performance so poor even after the fix? 7000 records/s is abysmal. That's still 2000-3000 times slower than the other benchmarks in this file.

dawidwys: Please remember the test time includes job submission. Moreover the results are returned via an iterator over an RPC/HTTP from the jobmaster. The test is not comparable in any sense to all the others.

dawidwys: If you have concerns towards the benchmark itself, I suggest merging the fix and create a separate followup for the benchmark.

pnowojski: With throughput so low, you are not benchmarking any performance critical code path. I'm not sure what's causing this problems, maybe indeed collecting the results, but whatever it is, it's the performance bottleneck, which your benchmark is benchmarking. And if collecting results via HTTP request is that bottleneck, that's definitely not what we want to benchmark. 

On the other hand, something might be still very wrong with this code, and the performance regression might be still out there. Without properly working benchmark how do you know you have fixed the performance regression?

src/main/java/org/apache/flink/benchmark/MultipleInputBenchmark.java
```diff
@@ -82,10 +98,86 @@ public void multiInputOneIdleMapSink(FlinkEnvironmentContext context) throws Exc
...
+		IdlingSource.reset();
+		try (final CloseableIterator<Long> iterator = stream.executeAndCollect()) {
+			while (iterator.hasNext()) {
+				final Long next = iterator.next();
+				if (next == CHAINED_IDLE_RECORDS_PER_INVOCATION) {
+					IdlingSource.signalCanFinish();
+				}
+			}
+		}
+	}
```
pnowojski, line 138:
```suggestion
        env.addOperator(transform);

        IdlingSource.reset();
        new MultipleConnectedStreams(env)
                .transform(transform)
                .addSink(new SinkClosingIdlingSource())
                .setParallelism(1);
        context.execute();
    }

    private static class SinkClosingIdlingSource implements SinkFunction<Long> {
        private int recordsSoFar = 0;

        @Override
        public void invoke(Long value) throws Exception {
            if (++recordsSoFar >= CHAINED_IDLE_RECORDS_PER_INVOCATION) {
                IdlingSource.signalCanFinish();
            }
        }
    }
```

pnowojski: with this change this benchmark starts working normally and you can safely increase number of records to 15M or 25M.

And only with this change you can see the true scope of the performance regression.
without fix ~7_000 records/s
wit fix ~14_000_000records/s

dawidwys: At least in my setup I still can not run it with more than 5 digit number of records without the fix.

pnowojski (approved):

PR #32, opened 2021-09-24
[FLINK-24230] Buffer debloating microbenchmark for single gate

dawidwys:
Added a test that first does an environment setup.
1. Submits a job
2. Waits for it being backpressured
3. Waits some extra time for buffer debloating to stabilize
The above steps are not measured. After that we trigger a single
checkpoint and measure the time for it to complete.


src/main/java/org/apache/flink/benchmark/BufferDebloatedCheckpointTimeBenchmark.java
```diff
@@ -0,0 +1,261 @@
...
+			JobVertexBackPressureInfo responseBody;
+			do {
+				responseBody = restClient.sendRequest(
+						restAddress.getHost(),
+						restAddress.getPort(),
+						JobVertexBackPressureHeaders.getInstance(),
+						metricsParameters,
+						EmptyRequestBody.getInstance()).get();
+			} while (responseBody.getBackpressureLevel() != JobVertexBackPressureInfo.VertexBackPressureLevel.HIGH);
+		}
```
pnowojski, line 213: Maybe add a timeout? I don't think we have any watchdog in benchmarks.

src/main/java/org/apache/flink/benchmark/BufferDebloatedCheckpointTimeBenchmark.java
```diff
@@ -0,0 +1,261 @@
...
+		checkpoint.get();
+	}
+
+	@State(Scope.Thread)
+	public static class DebloatedCheckpointEnvironmentContext {
+		public JobID jobID;
+		public StreamExecutionEnvironment env;
+		public MiniCluster miniCluster;
+
+		@Setup(Level.Invocation)
```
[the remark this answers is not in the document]
dawidwys: I used `Level.Invocation` so that the buffer debloating happens before each benchmark method. It's the matter of what we test in each benchmark.

Initially I thought we wanted to have tens of buffer debloating processes. However if you think we should test tens of checkpoints after a single buffer debloating I am good with that as well. It actually makes sense.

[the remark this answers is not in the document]
pnowojski: The purpose of the benchmark is to check effectiveness of the debloating, meassured as aligned checkpointing time. Multiple iterations/forks/invocations are only a mean to provide as stable results as possible. As long as results are trustworthy and stable, I don't mind if the debloating/setup happens once per fork (`Trial`), or once per invocation. I think the former one will provide more stable results, as we will even have a couple of "warm up" checkpoints before the measurement starts.

src/main/java/org/apache/flink/benchmark/BufferDebloatedCheckpointTimeBenchmark.java
```diff
@@ -0,0 +1,261 @@
...
+				new OptionsBuilder()
+						.verbosity(VerboseMode.NORMAL)
+						.include(BufferDebloatedCheckpointTimeBenchmark.class.getCanonicalName())
+						.build();
+
+		new Runner(options).run();
+	}
+
+	@Benchmark
+	public void unalignedCheckpoint(DebloatedCheckpointEnvironmentContext context)
```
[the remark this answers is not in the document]
dawidwys: I went with `debloatedCheckpointSingleInput`

src/main/java/org/apache/flink/benchmark/BufferDebloatedCheckpointTimeBenchmark.java
```diff
@@ -0,0 +1,261 @@
...
+		value = 3,
+		jvmArgsAppend = {
+				"-Djava.rmi.server.hostname=127.0.0.1",
+				"-Dcom.sun.management.jmxremote.authenticate=false",
+				"-Dcom.sun.management.jmxremote.ssl=false",
+				"-Dcom.sun.management.jmxremote.ssl"
+		})
+@Warmup(iterations = 4)
+@Measurement(iterations = 10)
+public class BufferDebloatedCheckpointTimeBenchmark {
```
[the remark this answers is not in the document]
dawidwys: How come? Do you mean if I could run the `BufferDebloatedCheckpointTimeBenchmark` with unaligned checkpoints enabled but buffer debloating disabled?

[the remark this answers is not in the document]
pnowojski: >  Do you mean if I could run the BufferDebloatedCheckpointTimeBenchmark with unaligned checkpoints enabled but buffer debloating disabled?

Yes % renames.  This way of benchmarking checkpoint times under backpressure seems just like a better idea compared to the old `UnalignedCheckpointTimeBenchmark`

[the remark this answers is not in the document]
dawidwys: Ok, I'll try.

dawidwys:
@pnowojski I think I unified the two tests. I have two concerns though:
1. Shall I use the old record size = 1kb or should I use the same record size as for debloating? With 64b records an unalgined/alternating checkpoint takes ~3s.
2. I can't see a difference in performance between the ALTERNATING(timeout=1) and UNALIGNED(timeout=0). In the old benchmark the difference was visible. Can you spot why is that? The results are:
```
CheckpointingTimeBenchmark.checkpointSingleInput                                  UNALIGNED  sample  120  311.043 ± 2.717  ms/op
CheckpointingTimeBenchmark.checkpointSingleInput                                ALTERNATING  sample  114  315.690 ± 4.936  ms/op
```

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
```diff
@@ -97,43 +100,119 @@
...
+                    config.set(TaskManagerOptions.MEMORY_SEGMENT_SIZE, START_MEMORY_SEGMENT_SIZE);
+                    config.set(
+                            ExecutionCheckpointingOptions.ALIGNED_CHECKPOINT_TIMEOUT,
+                            Duration.ofMillis(0));
+                    config.set(TaskManagerOptions.BUFFER_DEBLOAT_ENABLED, false);
+                    return config;
+                },
+                () -> null,
+                UNALIGNED_RECORD_SIZE),
+        ALTERNATING(
```
pnowojski, line 142: Why do you call it `ALTERNATING`?

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
```diff
@@ -97,43 +100,119 @@
...
+                },
+                () -> {
+                    // wait a bit for debloating to stabilize
+                    Thread.sleep(2_000);
+                    return null;
+                },
+                DEBLOATING_RECORD_SIZE);
+
+        private final Function<Configuration, Configuration> configFunc;
+        private final SupplierWithException<Void, Exception> postSetUpFunc;
```
pnowojski, line 175: `FunctionWithException`? Or even better, why don't you change it to a simple `int setupSleepDelay`?

pnowojski (commented):
> I can't see a difference in performance between the ALTERNATING(timeout=1) and UNALIGNED(timeout=0). In the old benchmark the difference was visible. Can you spot why is that? The results are:

The results themselves are also strange, checkpointing is much slower compared to the old benchmarks. I haven't spotted anything obvious, but those numbers should be roughly in sync with the existing benchmark with the same configuration. If they are not, we should understand why they are not.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
dawidwys: Isn't that the differrence between timeout = 0 and timeout = 1?

dawidwys:
There are two significant difference to the old benchmark:
* the new one is run with parallelism = 4 (old one was run with 1)
* we measure checkpoints only when there is high backpressure, in the old we don't really know to what extent the pipeline is saturated
I am not sure if we can compare those results.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
dawidwys: I don't like the `int setupSleepDelay` because it is less flexible, plus I'd need to add an `if (setupSleepDelay > 0)`.

What's wrong with `Supplier`? It's just a function that does not take arguments.

pnowojski: > I don't like the int setupSleepDelay because it is less flexible, plus I'd need to add an if (setupSleepDelay > 0).

You could have the same argument in favour of replacing all of the regular function arguments with some lambda functions. Yes, it's less flexibly, but why use more complicated construct for something that could be expressed easily as a single parameter?

> What's wrong with Supplier? It's just a function that does not take arguments.

Readability. Putting aside strange `return null;` that does nothing, in order to understand what is the following line doing:
```
mode.postSetUp();
```
With supplier/function you need to jump through the code much more and this flexibility is actually hurting your understanding, as you need to see a definition of every possible `SupplierWithException<Void, Exception> postSetUpFunc` parameter to understand what is this single line doing in all of the cases. Currently, in this case, it's not that big of a deal, as everything is one file, relatively close to one another, but the point still stands. If you see
```
mode.postSetUp() {
  postSetUpFunc.get();
}
```
it's very non obvious what is this doing. With
```
Thread.sleep(mode.postSetUpSleep);
```
it's much easier to understand.

pnowojski:
> the new one is run with parallelism = 4 (old one was run with 1)

Can you compare the results with parallelism = 1?

> we measure checkpoints only when there is high backpressure, in the old we don't really know to what extent the pipeline is saturated

The old benchmark is doing many checkpoints, and the configuration was supposed to be tweaked in such a way, that the job should back pressure very very quickly, within first or a couple of first checkpoints. So the average number of checkpoints/s from old benchmark should be comparable to yours one show invocation time.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
dawidwys: That's the name that is used all across the code base (`AlternatingCollectingBarriers`, `AlternatingWaitingForFirstBarrier`, `org.apache.flink.streaming.runtime.io.checkpointing.SingleCheckpointBarrierHandler#alternating`) exactly for the situation you described. 

Checkpoints are alternating between aligned and unaligned.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
dawidwys: I will not argue about such a minimal change... Will update it, even if I see it differently.

dawidwys:
Ok, I found the issue with unaligned checkpoints. They were dropped silently because the checkpointing was not enabled. There is a check in `org.apache.flink.streaming.api.graph.StreamingJobGraphGenerator#getCheckpointingMode` which switches the Checkpointing mode to `AT_LEAST_ONCE` which in turn disables unaligned checkpoints (`StreamingJobGraphGenerator:265).

Unforunately it's not that perfect, as the `ALIGNED` checkpoint do not match the math anymore. Or do they?

```
Benchmark                                                                            (mode)    Mode   Cnt     Score    Error  Units
CheckpointingTimeBenchmark.checkpointSingleInput                                    ALIGNED  sample    19   878.707 ± 71.026  ms/op
CheckpointingTimeBenchmark.checkpointSingleInput                                  UNALIGNED  sample  4051     2.465 ±  0.080  ms/op
CheckpointingTimeBenchmark.checkpointSingleInput                                ALTERNATING  sample    82   132.851 ±  6.456  ms/op
```

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
pnowojski: Ok, I understand your confusion. No, the checkpoints are unaligned, maybe with or without time out. `Alternating` relates to how its implemented internally, mostly in order to support aligned savepoints, but yes, also unaligned checpkoint barriers with timeout are implemented using aligned checkpoint barriers. But that's from the perspective of the given `CheckpointBarrierHandler`, which basically can expect aligned/unaligned checkpoint barriers arriving to it. From the perspective of a job/user, there is no alternation. There are just unaligned checkpoints with different timeout values. "Alternating" term is not used anywhere in the docs/public API/configs.

In other words, it's an implementation detail of how `CheckpointBarrier` is implemented. It could have been implemented differently, for example by squashing the unaligned and aligned `CheckpointBarrierHandler` code into a single `UnaligndAndAlignedCheckpointBarrierHandler` that could support both aligned and unaligned checkpoints. However that would duplicate the code of the aligned version of the `CheckpointBarrierHandler`, we it's implemented as a `CheckpointBarrierHandler` that can alternate between aligned/unaligned barriers.

dawidwys: Cutting the discussion short ;) Any ideas for better names? I don't think TIMEOUT_0, TIMEOUT_1 are more informative.

pnowojski:
> Unforunately it's not that perfect, as the ALIGNED checkpoint do not match the math anymore. Or do they?

You can run the benchmark with different debloating target to check how it's behaving, but 878ms vs 600ms from the debloating target is not that far off. Maybe a couple of ms are wasted on the initial RPCs/CheckpointCoordinator/sync/async checkpoint phases. Maybe you could check the checkpoint statistics somehow?

Furthermore, [there is a 50% threshold for buffer adjustments](https://ci.apache.org/projects/flink/flink-docs-master/docs/deployment/config/#taskmanager-network-memory-buffer-debloat-threshold-percentages), so if estimated time to consume the data is 430ms and we would like to reduce it down to 300ms, that's below the threshold value and the new buffer size won't be announced. Having said that, maybe it's a good idea to reduce this threshold to a couple of % for our micro benchmark (more frequent buffer size announcements won't matter much in this case). Maybe it would also help to bring down the noise/`Error` of that benchmark? `± 71.026` is quite a lot, and maybe it's caused by different invocations  being stuck on different thresholds?

dawidwys:
I tried changing the debloating target with basically a similar outcome. Changing the threshold also doesn't seem to help and does not have any visible effect. After some debugging the buffer size just jumps back and forth a lot, irrespective of the setting.

I added logs in `BufferDebloater` when announcing the new size of buffers. I observed that in the setup from this PR, the throughput varied quite a lot. I tried increasing the `BUFFER_DEBLOAT_PERIOD` which helped with the throughput measurement, but did not solve the issue with unstable results (and too high results).

One thing I noticed and haven't understood and solved yet is that I see that the `totalNumberOfBuffersInUse` also varies a lot:
```
14:50:44,461 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 1964. Throughput: 26195, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 4
14:50:44,462 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 1230. Throughput: 32804, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 8
14:50:44,763 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 303. Throughput: 25314, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 25
14:50:44,764 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 527. Throughput: 31636, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 18
14:50:44,765 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 312. Throughput: 29151, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 28
14:50:44,784 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 874. Throughput: 26238, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 9
14:50:44,785 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 786. Throughput: 26225, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 10
14:50:44,785 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 873. Throughput: 26212, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 9
14:50:44,788 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 1574. Throughput: 26246, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 5
14:50:45,064 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 1812. Throughput: 30215, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 5
14:50:45,067 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 2864. Throughput: 47740, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 5
14:50:45,086 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 263. Throughput: 26333, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 30
14:50:45,087 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 282. Throughput: 26368, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 28
14:50:45,087 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 256. Throughput: 26135, lastEstimatedTimeToConsumeBuffers: PT0.333S, totalNumber: 34
14:50:45,091 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 442. Throughput: 26572, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 18
14:50:45,366 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 295. Throughput: 28591, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 29
14:50:45,369 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 1031. Throughput: 44710, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 13
14:50:45,668 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 321. Throughput: 31108, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 29
14:50:45,691 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 600. Throughput: 26033, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 13
14:50:45,691 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 561. Throughput: 26181, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 14
14:50:45,692 ERROR org.apache.flink.streaming.runtime.tasks.bufferdebloat.BufferDebloater [] - Announcing new buffer size: 522. Throughput: 26139, lastEstimatedTimeToConsumeBuffers: PT0.299S, totalNumber: 15
```
which causes the buffer size to jump back and forth.

akalash:
I compared this benchmark with UnalignedCheckpointTimeBenchmark and they are perfectly the same. Just in case, it is not possible to compare it directly because this benchmark shows pure checkpoint time(because we triggered it manually) while the old one has an overhead like the checkpoint interval and so on. But if we take the average pure time from the old benchmark it would be the same as in the new one.
According to the buffer debloat, I don't see the big problem here, I mean it is indeed volatile but it is exactly how it works now especially with such a small number in configuration. I tried many other configurations but the result is the same. So I propose to keep this benchmark as is and then check the trend after a while.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,301 @@
...
+        public int getSleepPostSetUp() {
+            return sleepPostSetUp;
+        }
+    }
+
+    @State(Scope.Thread)
+    public static class DebloatedCheckpointEnvironmentContext extends FlinkEnvironmentContext {
+        public JobID jobID;
+
+        @Param({"ALIGNED", "UNALIGNED", "ALTERNATING"})
```
pnowojski, line 201: bump `ALTERNATING` issue. Rename to `TIMEOUT_1`? `UNALIGNED_1`?

pnowojski (changes requested):
One more thing. I would try to keep the reported value as "checkpoints/s" or "checkpoints/ms", to have consistence "more is better" in the benchmarks/charts. Otherwise it would be confusing and it would require manually changing codespeed settings to mark "less is better" for all of those benchmarks.

src/main/java/org/apache/flink/benchmark/CheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,291 @@
...
+ *       to be processed by the sink
+ *   <li>We have 2 (exclusive buffers) * 4 (parallelism) + 8 floating = 64 buffers per gate, with
+ *       300 ms debloating target and ~1ms/record processing speed, we can buffer 300/64 = ~4.5
+ *       records in a buffer after debloating which means the size of a buffer is slightly above the
+ *       minimal memory segment size.
+ *   <li>The buffer debloating target of 300ms means a checkpoint should take ~2(number of
+ *       exchanges)*300ms=~600ms
+ * </ul>
+ */
+@OutputTimeUnit(SECONDS)
```
pnowojski, line 91: Ahhh, sorry for one more unnecessary round, but having `ops/ms` would be even better, as that's the default unit in the codespeed on the charts. (actually it's `records/ms` but let's loosely define "record" as a checkpoint barrier ;) )

dawidwys: I don't think `ops/ms` is a good idea. For e.g. `ALIGNED` case (~600-800ms/checkpoint) it goes below 0.001 op/ms, which I think is the lowest granularity for jmh (unless there is some magic setting I am not aware of).

pnowojski (approved):
LGTM,  let's see how will it behave on benchmarking machine over the next couple of days.

PR #35, opened 2021-10-07
[FLINK-24231] Benchmark for buffer debloating multiple gates 

dawidwys:
Based on #34

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.benchmark.operators.RecordSource;
+import org.apache.flink.benchmark.operators.RecordSource.Record;
+import org.apache.flink.configuration.MemorySize;
+import org.apache.flink.runtime.jobgraph.JobGraph;
+import org.apache.flink.runtime.jobgraph.JobVertex;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.functions.co.CoMapFunction;
+import org.apache.flink.streaming.api.graph.StreamGraph;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+@OutputTimeUnit(SECONDS)
+public class MultiInputCheckpointingTimeBenchmark extends BenchmarkBase {
+
+    public static final int JOB_PARALLELISM = 4;
+    public static final MemorySize START_MEMORY_SEGMENT_SIZE = MemorySize.parse("8 kb");
+    public static final MemorySize MIN_MEMORY_SEGMENT_SIZE = MemorySize.parse("64 b");
+    public static final Duration DEBLOATING_TARGET = Duration.of(300, ChronoUnit.MILLIS);
+    public static final MemorySize SMALL_RECORD_SIZE = MemorySize.parse("1b");
+    public static final MemorySize BIG_RECORD_SIZE = MemorySize.parse("1kb");
+    public static final int DEBLOATING_STABILIZATION_PERIOD = 2_000;
```
pnowojski, line 73: Shouldn't we extract those constants to a common base class?

src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+            final List<JobVertex> vertices = jobGraph.getVerticesSortedTopologicallyFromSources();
+
+            final JobClient jobClient = env.executeAsync();
+            jobID = jobClient.getJobID();
+            CommonTestUtils.waitForAllTaskRunning(miniCluster, jobID, false);
+            List<JobVertexID> sourceIds =
+                    Arrays.asList(vertices.get(0).getID(), vertices.get(1).getID());
+            BackpressureUtils.waitForBackpressure(
+                    jobID, sourceIds, miniCluster.getRestAddress().get());
+            Thread.sleep(DEBLOATING_STABILIZATION_PERIOD);
```
pnowojski, line 150: This I think we could also deduplicate quite easily in the base class?

src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+     * The custom sink for processing records slowly to cause accumulate in-flight buffers even back
+     * pressure.
+     */
+    public static class SlowDiscardSink<T> implements SinkFunction<T> {
+        @Override
+        public void invoke(T value, Context context) {
+            final long startTime = System.nanoTime();
+            while (System.nanoTime() - startTime < 200_000) {}
+        }
+    }
```
pnowojski, line 188: ditto

src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+            final Configuration config = super.createConfiguration();
+            config.set(ExecutionCheckpointingOptions.ENABLE_UNALIGNED, false);
+            config.set(TaskManagerOptions.MEMORY_SEGMENT_SIZE, START_MEMORY_SEGMENT_SIZE);
+            config.set(TaskManagerOptions.MIN_MEMORY_SEGMENT_SIZE, MIN_MEMORY_SEGMENT_SIZE);
+            config.set(TaskManagerOptions.BUFFER_DEBLOAT_ENABLED, true);
+            config.set(TaskManagerOptions.BUFFER_DEBLOAT_TARGET, DEBLOATING_TARGET);
+            config.set(
+                    TaskManagerOptions.BUFFER_DEBLOAT_PERIOD, Duration.of(200, ChronoUnit.MILLIS));
+            config.set(TaskManagerOptions.BUFFER_DEBLOAT_SAMPLES, 5);
+            return config;
```
pnowojski, line 164: Maybe even this? Isn't it the same as `ALIGNED` mode from the singleInput benchmark?

src/main/java/org/apache/flink/benchmark/MultiInputCheckpointingTimeBenchmark.java
pnowojski: I understand what you mean. If we happen to need to tweak the parameters differently in those two benchmark, I think we should just then have such parameters defined separately in two files. I think we can achieve this quite easily via some abstract getters in the context class?

However if they are now the same, I would like to avoid a situation when someone in the future changes a parameter only in one file, but he simply forgets to change/check it in the other one, causing lot's of confusion for next one in line that will notice the unnecessary difference.

What I'm trying to say, is that ideally we should come up with a code structure that reflects both of those aspects - a shared parameters and private ones (if we need them), which would give us more self documenting code.

PR #37, opened 2021-10-13
[FLINK-24531] Fix connection time out when checking the backpressure

pnowojski:


src/main/java/org/apache/flink/benchmark/BackpressureUtils.java
```diff
@@ -65,20 +69,27 @@ static void waitForBackpressure(JobID jobID, List<JobVertexID> sourceId, URI res
         }
     }
 
+    private static RestClusterClient<StandaloneClusterId> clientConfiguration(
```
dawidwys, line 72: nit: `createClient`?

PR #39, opened 2021-11-01
[FLINK-24659] Drop RemoteBenchmarkBase.miniCluster field

pnowojski:
Before this change all children of RemoteBenchmarkBase worked incorrectly since they
configured the environment for miniCluster from FlinkEnvironmentContext but in reality,
they used miniCluster from RemoteBenchmarkBase.


akalash (approved):
@pnowojski, LGTM

PR #41, opened 2021-11-21
[FLINK-24918][Runtime/State Backends]Support to specify the data dir for state benchmark

Aitozi:
Rely on the Flink [PR](https://github.com/apache/flink/pull/17846)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ mvn clean package exec:exec \
 ```
 
 If you want to execute just one benchmark, the best approach is to execute selected main function manually.
-There're mainly two ways:
+There're mainly three ways:
 
 1. From your IDE (hint there is a plugin for Intellij IDEA).
    * In this case don't forget about selecting `flink.version`, default value for the property is defined in pom.xml.
@@ -31,13 +31,37 @@ There're mainly two ways:
    mvn -Dflink.version=<FLINK_VERSION> clean package exec:exec \
     -Dbenchmarks="<benchmark_class>"
    ```
-An example flink version can be -Dflink.version=1.12-SNAPSHOT.
+
+    An example flink version can be -Dflink.version=1.12-SNAPSHOT.
+
+3. Run the uber jar directly like:
+
+    ```
+    java -jar target/benchmarks.jar -rf csv "<benchmark_class>"
+    ```
 
 We also support to run each benchmark once (with only one fork and one iteration) for testing, with below command:
 
 ```
 mvn test -P test
 ```
+
+## Parameters
+
+There are some built-in parameters to run different benchmarks, these can be shown/overridden from the command line.
+
+```
+# show all the parameters combination for the <benchmark_class> 
+java -jar target/benchmarks.jar "<benchmark_class>" -lp
+
+# run benchmark for rocksdb state backend type 
+java -jar target/benchmarks.jar "org.apache.flink.state.benchmark.*" -p "backendType=ROCKSDB" 
+```
+
+## Configuration
+
+Besides the parameters, there is also a benchmark config file `benchmark-conf.yaml` to tune some basic parameters. 
+For example, we can change the state data dir by putting `benchmark.state.data-dir: /data` in the config file. For more options, you can refer to the code in the `org.apache.flink.config` package. 
 
 ## Prerequisites
 
--- a/src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
+++ b/src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
@@ -39,7 +39,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils.applyToAllKeys;
 import static org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils.compactState;
-import static org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils.createKeyedStateBackend;
 import static org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils.getListState;
 import static org.apache.flink.state.benchmark.StateBenchmarkConstants.listValueCount;
 import static org.apache.flink.state.benchmark.StateBenchmarkConstants.setupKeyCount;
@@ -64,7 +63,7 @@ public class ListStateBenchmark extends StateBenchmarkBase {
 
     @Setup
     public void setUp() throws Exception {
-        keyedStateBackend = createKeyedStateBackend(backendType);
+        keyedStateBackend = createKeyedStateBackend();
         listState = getListState(keyedStateBackend, STATE_DESC);
         dummyLists = new ArrayList<>(listValueCount);
         for (int i = 0; i < listValueCount; ++i) {
--- a/src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
+++ b/src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
@@ -18,6 +18,9 @@
 package org.apache.flink.state.benchmark;
 
 import org.apache.flink.benchmark.BenchmarkBase;
+import org.apache.flink.config.ConfigUtil;
+import org.apache.flink.config.StateBenchmarkOptions;
+import org.apache.flink.configuration.Configuration;
 import org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils;
 import org.apache.flink.runtime.state.KeyedStateBackend;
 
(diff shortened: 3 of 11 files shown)


Aitozi:
This build error due to it reply on the new api in [PR](https://github.com/apache/flink/pull/17846)

README.md
```diff
@@ -31,14 +31,33 @@ There're mainly two ways:
    mvn -Dflink.version=<FLINK_VERSION> clean package exec:exec \
```
pnowojski, line 31: nitty nit: In line 24 replace:
> There're mainly two way

with

> There're three ways

?

README.md
```diff
@@ -31,14 +31,33 @@ There're mainly two ways:
...
+    java -jar target/benchmarks.jar -rf csv "<benchmark_class>"
+    ```
 
 We also support to run each benchmark once (with only one fork and one iteration) for testing, with below command:
 
 ```
 mvn test -P test
 ```
 
+## Params
```
pnowojski, line 49: nit: (here and in other places) `params` -> `parameters`

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
```diff
@@ -52,10 +56,21 @@
     final ThreadLocalRandom random = ThreadLocalRandom.current();
 
     @Param({"HEAP", "ROCKSDB"})
-    protected StateBackendBenchmarkUtils.StateBackendType backendType;
+    private StateBackendBenchmarkUtils.StateBackendType backendType;
+
+    @Param({"/tmp"})
```
pnowojski, line 61: Will this be behaving in the same/similar way as it used to be? What if on the given machine there is no `/tmp` directory or the JVM process doesn't have an access to it? I think the old `null` value was more resilient?

pnowojski (commented):
+1 for the change. I've left a couple of comments.

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
Aitozi: It's indeed have different behavior, because it's depends on the `java.io.tmpdir` property. 
The `@param` do not accept `null`, so I use `not_set` as default value . 

```
# Warmup: 10 iterations, 1 s each
# Measurement: 10 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.apache.flink.state.benchmark.ValueStateBenchmark.valueAdd
# Parameters: (backendType = HEAP, stateDataDirPath = not_set)
```

Aitozi:
@pnowojski  Thanks for your review, I have addressed your comments.

README.md
```diff
@@ -31,14 +31,33 @@ There're mainly two ways:
...
 
 We also support to run each benchmark once (with only one fork and one iteration) for testing, with below command:
 
 ```
 mvn test -P test
 ```
 
+## Parameters
+
+There are some built-in params to run different benchmarks, these can be shown/overridden from the command line.
```
pnowojski, line 51: Missed `params`

README.md
```diff
@@ -31,14 +31,33 @@ There're mainly two ways:
...
+
+```
+# show all the params combination for the <benchmark_class>
```
pnowojski, line 54: Missed `params`

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
```diff
@@ -26,6 +26,7 @@
 import org.openjdk.jmh.annotations.Level;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.infra.BenchmarkParams;
```
pnowojski, line 29: Why is this needed?

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
```diff
@@ -52,10 +54,25 @@
     final ThreadLocalRandom random = ThreadLocalRandom.current();
 
     @Param({"HEAP", "ROCKSDB"})
-    protected StateBackendBenchmarkUtils.StateBackendType backendType;
+    private StateBackendBenchmarkUtils.StateBackendType backendType;
+
+    @Param({StateBenchmarkConstants.defaultStateDataDirPath})
+    private String stateDataDirPath;
```
pnowojski, line 60: Won't this in combination with https://github.com/apache/flink-benchmarks/blob/master/save_jmh_result.py#L72:L76 change the names of the uploaded benchmarks to the [code speed](http://codespeed.dak8s.net:8000/changes/)?

Do we really want this to be reported like this? Won't this clog the results and the WebUI in particular?
Secondly, even if we want this, we would have to manually rename all of the benchmarks in the database to preserve continuity of the results in the graphs.

Maybe it's actually better idea after all, to not use jmh's `@Param` for this? Maybe simply use an environmental variable?

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
```diff
@@ -20,15 +20,17 @@
 import org.apache.flink.benchmark.BenchmarkBase;
 import org.apache.flink.contrib.streaming.state.benchmark.StateBackendBenchmarkUtils;
 import org.apache.flink.runtime.state.KeyedStateBackend;
-
```
pnowojski, line 23: Are you using the same formatting settings as for the main Flink repository? 

https://nightlies.apache.org/flink/flink-docs-master/docs/flinkdev/ide_setup/#code-formatting

pnowojski (commented):
Thanks for the updates. I've left a couple of minor comments, and one more issue that I've just realised about.

src/main/java/org/apache/flink/state/benchmark/ListStateBenchmark.java
Aitozi: removed

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
Aitozi: updated

src/main/java/org/apache/flink/state/benchmark/StateBenchmarkBase.java
Aitozi: I didn't notice it. Now I prefer to not use `@Param` for this . Instead of using environmental variable, Can we add a benchmark conf as I mentioned in [FLINK-24967](https://issues.apache.org/jira/projects/FLINK/issues/FLINK-24967)

pnowojski: I think either of those options would work, depends on what we want to achieve? I think using `@Params` makes sense if we really want to present the results from different configurations side by side, especially in the WebUI or in the continuous regression testing scripts. However if there are many many different constants that you would like to change, it probably doesn't make sense to keep them all separately as `@Params`, as it would simply explode the number of tests (jmh is testing all combinations of params). 

Having said all of that, a dedicated config file might be a not a bad solution. For now there can be a single configuration file that anyone can overwrite as he chooses. If we ever decide to present the results from different configs side by side, we can do this via `@Param` `preset`, where `preset` would be a name of one of those config files.

Aitozi: ok, I will provide a single configuration implementation .

Aitozi:
@pnowojski Sorry for the late response, I have done the work to add support for a dedicated config file. Please take a look when you are free. 

I have check the ci failure log , It seems unrelated to this Pr.

```
 Could not find artifact org.apache.flink:flink-examples-batch_2.12:jar:tests:1.15-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
```

pnowojski (approved):
LGTM % two nits

Aitozi:
Thanks for your comments @pnowojski , I have updated the PR accordingly

pnowojski (approved):
Thanks for the update. LGTM,. Can you squash the 2nd, 3rd and 4th commits together? Do you have committer writes and can you merge this PR on your own, or should I do it?

Aitozi:
@pnowojski I squashed the commit, I do not have committer writers, please help to merge it 😄

pnowojski:
Thanks @Aitozi. I will do it, but could you take a look at the travis failure? It looks probably unrelated to your change, but it blocks merging of this PR.

Aitozi:
OK, I will try to look into it tonight.

Aitozi:
It should be caused by [[FLINK-25039][build] Disable shading of test jar by default]( https://github.com/apache/flink/pull/17895 ). After that, the `flink-examples-batch_2.12` will not generate an empty test-jar. 

Since the test-jar is empty, so we do not depend on it actually. I think it can be removed safely. Let's see will the travis turn green :)

pnowojski:
Thanks! LGTM. Can you just rename the last commit by prefixing it with `[FLINK-25039]`?

Aitozi:
done @pnowojski

pnowojski:
Thanks :) merging

PR #44, opened 2022-01-26
[FLINK-25704] Fix the blocking partition benchmark regression caused by FLINK-25637

wsry:
FLINK-25637 changed the default blocking shuffle implementation from hash-shuffle to sort-shuffle which caused some benchmark regression. This PR tries to fix the regression. For more information about the regression, please refer to https://issues.apache.org/jira/browse/FLINK-25704.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/BlockingPartitionBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/BlockingPartitionBenchmark.java
@@ -95,6 +95,9 @@ public class BlockingPartitionBenchmark extends BenchmarkBase {
                 boolean compressionEnabled, String subpartitionType) {
             Configuration configuration = super.createConfiguration();
 
+            configuration.setInteger(
+                    NettyShuffleEnvironmentOptions.NETWORK_SORT_SHUFFLE_MIN_PARALLELISM,
+                    Integer.MAX_VALUE);
             configuration.setBoolean(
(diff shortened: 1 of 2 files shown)


pnowojski (commented):
Hey @wsry. Can you explain how this PR is trying to fix the regression? How changing `NETWORK_SORT_SHUFFLE_MIN_PARALLELISM` is supposed to:
> Make the existing benchmark tests still test BoundedBlockingResultPartition;

?

wsry:
> NETWORK_SORT_SHUFFLE_MIN_PARALLELISM

@pnowojski Thanks for the review.

```NETWORK_SORT_SHUFFLE_MIN_PARALLELISM``` is a config option to config which blocking shuffle implementation to use, either the hash-based one or the sort-based one. If task parallelism is smaller than this config value, the hash-based blocking shuffle will be used, otherwise, the sort-based blocking shuffle will be used. Previously, the default value of this config option is Integer.MAX_VALUE, which mean hash-shuffle will be used by default, FLINK-25637 changed the default value to 1, which means sort-shuffle will be used by default. The changes in this PR restore the previous default value for the corresponding benchmark cases.

pnowojski (approved):
Thanks for the explanation. Could you maybe copy paste this to the commit message for future reference?

LGTM

wsry:
Updated commit message.

PR #50, opened 2022-03-16
[FLINK-26682] Add regression detection script

pnowojski:
Based upon https://github.com/rkhachatryan/flink-benchmarks/blob/regression-alerts/regression-alert.sh from @rkhachatryan .

### DIFF
--- a/regression_report.py
+++ b/regression_report.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


regression_report.py
```diff
@@ -0,0 +1,143 @@
...
+    if len(results) < args.baseLine:
+        return
+    median = getMedian(results[args.recentTrend:args.baseLine])
+    recent_median = getMedian(results[:args.recentTrend])
+    median_trend = 0
+    if median != 0:
+        median_trend = recent_median * 100 / median - 100
+    if median_trend < args.medianTrendThreshold:
+        # regression detected, print warning
+        print "<%s|%s> median=%s recent_median=%s" % (urlToBenchmark, benchmark, median, recent_median)
```
akalash, line 136: How or where should this script work? I mean I thought that we have jenkins job which periodically can check for regression and publish the result to the slack. But here I see only printing into the console. So what the main scenario of usage of this script?

akalash (approved):
LGTM

regression_report.py
pnowojski: As described offline. Jenkins is reading stdout from this script and pushing it to our monitoring system.

PR #56, opened 2022-06-24
[FLINK-28241] Overridden REQUIREMENTS CHECK DELAY to zero to avoid de…

akalash:
…lay on initial phase of each job start.

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/FlinkEnvironmentContext.java
+++ b/src/main/java/org/apache/flink/benchmark/FlinkEnvironmentContext.java
@@ -34,6 +34,9 @@ import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
 
+import java.time.Duration;
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/flink/benchmark/FlinkEnvironmentContext.java
```diff
@@ -107,6 +110,7 @@ protected Configuration createConfiguration() {
         // It doesn't make sense to wait for the final checkpoint in benchmarks since it only prolongs
         // the test but doesn't give any advantages.
         configuration.set(ExecutionCheckpointingOptions.ENABLE_CHECKPOINTS_AFTER_TASKS_FINISH, false);
+        configuration.set(REQUIREMENTS_CHECK_DELAY, Duration.ZERO);
```
rkhachatryan, line 113: nit: after FLINK-28243, we will likely remove this line. So I'd comment here in the code or mention it in FLINK-28243, or both.

rkhachatryan (approved):
Thanks for the fix @akalash, LGTM

PR #64, opened 2023-01-17
[FLINK-30717][ci] Migrate to GitHub Actions

zentol:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+          cache: 'maven'
+
+      - name: Set Maven 3.8.6
+        uses: stCarolas/setup-maven@v4.2
+        with:
+          maven-version: 3.8.6
+
+      - name: Install prerequisites
+        run: |
+          wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl1.0/libssl1.0.0_1.0.2n-1ubuntu5.10_amd64.deb
```
zentol, line 41: This is required for netty which only works with openssl 1.0. 
We could consider baking this into an image so we avoid the infrequent issues we have with the azure e2e tests where this link becomes outdated.

pnowojski: I presume that if this links is outdated, or the lib ssl is not installed properly, the tests will fail, and not that the issue will be silently ignored?

How likely is it that this version of the `libssl` will disappear? I mean how likely is it that baking this into the image will give us any benefit?

pnowojski (commented):
LGTM % one question

PR #66, opened 2023-02-27
[FLINK-29825] Replace median-based regression detect algorithm with max/min-based

fredia:
Replace median-based regression detect algorithm with max/min-based.
Detailed design is in https://docs.google.com/document/d/1Bvzvq79Ll5yxd1UtC0YzczgFbZPAgPcN3cI0MjVkIag/edit


fredia:
@pnowojski @lindong28 could you please take a look in free time? thanks!

lindong28 (commented):
Thanks for the PR! Left some comments below.

regression_report_v2.py
```diff
@@ -0,0 +1,108 @@
...
+    result = timelines['branches']['master'][exe]
+    lessIsbBetter = (timelines['lessisbetter'] == " (less is better)")
+    return result, lessIsbBetter
+
+def checkWithMax(urlToBenchmark, stds, scores, index, baselineSize):
+    sustainable_x = [min(scores[i - 2 : i + 1]) for i in range(index - baselineSize, index)]
+    baseline_throughput = max(sustainable_x)
+    current_throughput = max(scores[index - 3 : index])
+    current_unstable = stds[index] / current_throughput
+    if 1 - current_throughput / baseline_throughput > max(DEFAULT_THRESHOLD, 2 * current_unstable):
```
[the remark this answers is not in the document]
fredia: Replaced `DEFAULT_THRESHOLD` with `args. minRegressionRatio`, and added `minInstabilityMultiplier` to arguments.

regression_report_v2.py
```diff
@@ -0,0 +1,108 @@
...
+        checkWithMin(urlToBenchmark, stds, scores, len(scores) - 1,  args.baseLine)
+    else:
+        checkWithMax(urlToBenchmark, stds, scores, len(scores) - 1,  args.baseLine)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Regression report based on Max/Min value')
+    parser.add_argument('--base-line-size', dest='baseLine', required=False, default=DEFAULT_BASELINE, type=int,
+                        help='Number of samples taken as the base line.')
+    parser.add_argument('--download-samples-size', dest='downloadSamples', required=False, default=200,
```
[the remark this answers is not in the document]
fredia: > If we use a given number of commit points, then the baseline will be changed if there are a lot of commit points in a day.

Because the “commit-point ”(let's say run-point) of the benchmark is configured by Jenkins, it is assumed that a lot of commits are merged into the master in one day, but the benchmark only runs once on this day, so for a certain benchmark, there is only a run-point in this day.

Another reason is that codespeed does not provide a rest interface to fetch history data by day, so the number of run-points is used instead.

regression_report_v2.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if lessIsbBetter:
+        checkWithMin(urlToBenchmark, stds, scores, len(scores) - 1,  args.baseLine)
+    else:
+        checkWithMax(urlToBenchmark, stds, scores, len(scores) - 1,  args.baseLine)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Regression report based on Max/Min value')
+    parser.add_argument('--base-line-size', dest='baseLine', required=False, default=DEFAULT_BASELINE, type=int,
```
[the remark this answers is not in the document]
lindong28: I see. Thanks for the explanation.

regression_report_v2.py
```diff
@@ -0,0 +1,105 @@
...
+    urlToBenchmark = args.codespeedUrl + 'timeline/#/?' + urllib.urlencode({
+        'ben': benchmark,
+        'exe': exe,
+        'env': ENVIRONMENT,
+        'revs': args.displaySamples,
+        'equid': 'off',
+        'quarts': 'on',
+        'extr': 'on'})
+
+    if len(results) < args.baselineSize:
```
lindong28, line 72: Does this mean that a new benchmark will only be included in the regression detection 30 days after this benchmark is added?

I am wondering if it would be useful to include new benchmark in the detection faster. For example, we can start to detect regression after there are 5 samples for this benchmark.

regression_report_v2.py
```diff
@@ -0,0 +1,105 @@
...
+        scores = [-1 * score for score in scores]
+        direction = -1
+    detectRegression(urlToBenchmark, stds, scores, args.baselineSize, args.minRegressionRatio,
+                     args.minInstabilityMultiplier, direction)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Regression report based on Max/Min value')
+    parser.add_argument('--baseline-size', dest='baselineSize', required=False, default=30, type=int,
+                        help='Number of samples taken as the base line.')
```
lindong28, line 86: I am wondering if the following doc would be a bit more informative:

The maximum number of recent samples across which the maximum achieved throughput would be used as the baseline for regression detection.

regression_report_v2.py
```diff
@@ -0,0 +1,105 @@
...
+    parser.add_argument('--display-samples', dest='displaySamples', required=False, default=200,
```
lindong28, line 87: Since it refers to the number of samples, how about `numDisplaySamples`?

regression_report_v2.py
```diff
@@ -0,0 +1,105 @@
...
+    if lessIsbBetter:
+        scores = [-1 * score for score in scores]
+        direction = -1
+    detectRegression(urlToBenchmark, stds, scores, args.baselineSize, args.minRegressionRatio,
+                     args.minInstabilityMultiplier, direction)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Regression report based on Max/Min value')
+    parser.add_argument('--baseline-size', dest='baselineSize', required=False, default=30, type=int,
```
lindong28, line 85: Since it refers to the number of samples, how about naming it `numBaselineSamples`?

regression_report_v2.py
```diff
@@ -0,0 +1,105 @@
...
+import argparse
+import json
+import urllib
+import urllib2
+
+from regression_report import loadBenchmarkNames
+
+"""
+This is a regression detection algorithm based on the historical maximum/minimum value, please refer
+to https://docs.google.com/document/d/1Bvzvq79Ll5yxd1UtC0YzczgFbZPAgPcN3cI0MjVkIag/edit the detailed design.
```
lindong28, line 29: nits: could you help remove the word `edit` from this URL?

It might be useful to briefly describe the algorithm so that users would not have to visit the Google doc. How about the following doc:

```
"""
The regression detection algorithm calculates the regression ratio as the ratio of change between the current
throughput and the maximum throughput observed in the most recent numBaselineSamples samples. A regression alert is
triggered if the regression ratio exceeds max(minRegressionRatio, minInstabilityMultiplier * lastStandardDeviation)

Please refer to https://docs.google.com/document/d/1Bvzvq79Ll5yxd1UtC0YzczgFbZPAgPcN3cI0MjVkIag for more detail.
"""
```

lindong28 (commented):
Thanks @fredia for the update! LGTM overall. Left some minor comments.

BTW, do you have numbers showing how this algorithm performs in production? For example, is there false positive and false negative rate of this algorithm?

regression_report_v2.py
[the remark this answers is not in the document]
lindong28: I see. Thanks for the explanation.

regression_report_v2.py
fredia: > Does this mean that a new benchmark will only be included in the regression detection 30 days after this benchmark is added?

Yes, a new benchmark will only be included in the regression detection 30 days after this benchmark is added.
To detect new benchmark faster, It is a good idea to change the minimum sample limit to 5.

fredia: > Does this mean that a new benchmark will only be included in the regression detection 30 days after this benchmark is added?

Yes, a new benchmark will only be included in the regression detection 30 days after this benchmark is added.
To detect new benchmark faster, It is a good idea to change the minimum sample limit to 5.

regression_report_v2.py
fredia: 👍 After changing the minimum sample limit, this is a better description. Fixed.

fredia:
@lindong28  Thanks for the review and suggestions.

> do you have numbers showing how this algorithm performs in production? For example, is there false positive and false negative rate of this algorithm?

I have selected several benchmarks to test the precision and recall of this algorithm. Here are some data:

<div class="lake-content" typography="classic">

  | true positive | false positive | true negative | false negative | precision | recall
-- | -- | -- | -- | -- | -- | --
[arrayKeyBy](http://codespeed.dak8s.net:8000/timeline/#/?exe=1,5,8,9&ben=arrayKeyBy&env=2&revs=1000&equid=off&quarts=on&extr=on) | 42 | 2 | 840 | 10 | 0.96 | 0.83
[serializerRow_Java11](http://codespeed.dak8s.net:8000/timeline/#/?exe=5,6,8,9&ben=serializerRow&env=2&revs=1000&equid=off&quarts=on&extr=on) | 4 | 1 | 241 | 3 | 0.8 | 0.57
[serializerTuple_Java11](http://codespeed.dak8s.net:8000/timeline/#/?exe=5,6,8,9&ben=serializerTuple&env=2&revs=1000&equid=off&quarts=on&extr=on) | 4 | 24 | 218 | 3 | 0.14 | 0.57

Overall, It performs differently on different benchmarks, but its precision and recall can basically be greater than 50%.

Here is a [doc](https://docs.google.com/document/d/1coI4eJsauBtrlS1Z77bhGf-hNtDEXbzuwacG5ZPCMc8/edit?usp=sharing) comparing it to the current median-based algorithm, which contains their false positive and false negative data on different types of benchmarks.

PR #77, opened 2023-06-30
[FLINK-32506][connectors/common] Add the benchmark for watermark aggregation

1996fanrui:
https://issues.apache.org/jira/browse/FLINK-32506

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.apache.flink.benchmark;
+
+import org.apache.flink.runtime.source.coordinator.SourceCoordinatorAlignmentBenchmark;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+/** The watermark aggregation benchmark for source coordinator when enabling the watermark alignment. */
+public class WatermarkAggregationBenchmark extends BenchmarkBase {
+
+    private static final int NUM_SUBTASKS = 5000;
+
+    private static final int ROUND_PER_INVOCATION = 10;
+
+    private SourceCoordinatorAlignmentBenchmark benchmark;
+
+    public static void main(String[] args) throws RunnerException {
+        Options options =
+                new OptionsBuilder()
+                        .verbosity(VerboseMode.NORMAL)
+                        .include(".*" + WatermarkAggregationBenchmark.class.getCanonicalName() + ".*")
+                        .build();
+
+        new Runner(options).run();
+    }
+
+    @Setup(Level.Trial)
+    public void setup() throws Exception {
+        benchmark = new SourceCoordinatorAlignmentBenchmark();
+        benchmark.setup(NUM_SUBTASKS);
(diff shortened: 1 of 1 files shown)


1996fanrui:
CI fails due to https://github.com/apache/flink/pull/22921 isn't merged.

RocMarshal (approved):
hello, @1996fanrui nice benchmark patch~
LGTM +1(after https://github.com/apache/flink/pull/22921 merged.) on the whole and thank you for your effort.

src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+/** The watermark aggregation benchmark for source coordinator when enabling the watermark alignment. */
+@BenchmarkMode(Mode.AverageTime)
```
pnowojski, line 38: It would be great to keep the results consistent with other benchmarks, so as `ops/time`. `time/op` is "less is better", which causes later problems when trying to visualize/compare/analyse the results.

If you hardcode/inline `numSubtasks=5000` parameter, you could have `@BenchmarkMode(Throughput)` with `@OperationsPerInvocation(5000 * X)`. Where `X` is number of `sendReportedWatermarkToAllSubtasks` calls in a single `aggregateWatermark` invocation.

src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+public class WatermarkAggregationBenchmark extends BenchmarkBase {
+
+    @Param({"5000", "10000", "20000"})
```
pnowojski, line 41: I would run this benchmark only with a single parameter, to not prolong the benchmarking time. `5000` seems to be large enough to show problems, while also being somehow small enough to be reasonable value that could be expected in real world use cases.

src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+    public int numSubtasks;
+
+    @Param({"MONOTONE_INCREASING", "RANDOMIZE_MILLISECONDS"})
```
pnowojski, line 44: Does it matter? Can we drop this and have only a single hardcoded `RANDOMIZE_MILLISECONDS` parameter? Isn't `MONOTONE_INCREASING` unrealistic?

src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+    @Setup(Level.Trial)
+    public void setup() throws Exception {
+        benchmark = new SourceCoordinatorAlignmentBenchmark();
+        benchmark.setup(numSubtasks, timestampType);
+    }
+
+    @Benchmark
+    public void aggregateWatermark() {
+        benchmark.sendReportedWatermarkToAllSubtasks();
+    }
```
pnowojski, line 68: 3ms per invocation here might be on the border of being too small. It might make sense to have a loop of sending and reporting watermark 10 times per each `aggregateWatermark` invocation.

pnowojski (commented):
THanks!

src/main/java/org/apache/flink/benchmark/WatermarkAggregationBenchmark.java
pnowojski: huh, 918ops/ms ~= (5000/918) 5.44 ms/invocation with `NUM_SUBTASKS=5000`.

Previously you were reporting:
```
Benchmark                                         (numSubtasks)         (timestampType)  Mode  Cnt   Score   Error  Units
WatermarkAggregationBenchmark.aggregateWatermark           5000     MONOTONE_INCREASING  avgt   10   3.329 ± 0.070  ms/op
WatermarkAggregationBenchmark.aggregateWatermark           5000  RANDOMIZE_MILLISECONDS  avgt   10   3.541 ± 0.048  ms/op
```
why is the new version slower compared to the previous one? `5.44` vs  `3.54`? 🤔

1996fanrui: Hhh, I'm using 2 Macs, one is my work mac, one is my personal Mac. And my work mac is new version, its cpu is very strong than my Mac. I guess it's the toot cause 😀

1996fanrui: I can run this benchmark tomorrow on my work mac again.

1996fanrui: Following is the benchmark result of my work Mac:

```
The master branch with https://github.com/apache/flink/pull/22852
Benchmark                                          Mode  Cnt     Score    Error   Units
WatermarkAggregationBenchmark.aggregateWatermark  thrpt   10  1531.221 ± 45.514  ops/ms

The master branch
Benchmark                                          Mode  Cnt   Score   Error   Units
WatermarkAggregationBenchmark.aggregateWatermark  thrpt   10  17.743 ± 0.326  ops/ms
```

1531 ops/ms ~= (5000/1531) 3.26 ms/invocation, it's close and better than the previous one(3.54 ms/op). The little improvement may be  due to we  removed the `RANDOM` and `System.currentTimeMillis()` during benchmarking.

No matter which Mac or server, I am sure to see a huge performance improvement with https://github.com/apache/flink/pull/22852.

pnowojski (approved):
LGTM! (assuming green build, this should be merged after the changes to flink/master)

1996fanrui:
Thanks @pnowojski  and @RocMarshal  for the review, CI passed, merging~

Let's follow the benchmark result at http://codespeed.dak8s.net:8000/

PR #79, opened 2023-09-20
[FLINK-33122][benchmark] Support null checkpoint data path for rescaling benchmarks

Zakelly:
Currently, when setting up a rescaling benchmark, a local checkpoint storage is created based on a local path configured by "benchmark.state.data-dir". When user does not provide value for this option, an exception is thrown. In this case, the right behavior should be to create a temporary directory for checkpoint, just like the `StateBackendBenchmarkUtils#createKeyedStateBackend` does for local data directory.

### DIFF
--- a/src/main/java/org/apache/flink/state/benchmark/RescalingBenchmarkBase.java
+++ b/src/main/java/org/apache/flink/state/benchmark/RescalingBenchmarkBase.java
@@ -22,6 +22,8 @@ import org.apache.flink.api.common.state.ValueStateDescriptor;
 import org.apache.flink.api.common.typeinfo.PrimitiveArrayTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.benchmark.BenchmarkBase;
+import org.apache.flink.config.ConfigUtil;
+import org.apache.flink.config.StateBenchmarkOptions;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.state.benchmark.RescalingBenchmark;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/flink/state/benchmark/RescalingBenchmarkBase.java
```diff
@@ -57,6 +63,26 @@ public static void runBenchmark(Class<?> clazz) throws RunnerException {
         new Runner(options).run();
     }
 
+    protected static File prepareDirectory(String prefix) throws IOException {
```
masteryhx, line 66: All code paths is similar with StateBenchmarkBase#createKeyedStateBackend and StateBackendBenchmarkUtils#prepareDirectory, Could we merge them together ?

masteryhx (commented):
Thanks for the pr, PTAL my comments

src/main/java/org/apache/flink/state/benchmark/RescalingBenchmarkBase.java
Zakelly: Hi @masteryhx ,  The `StateBackendBenchmarkUtils` belongs to the Flink repo and the `prepareDirectory` is a private method. I'd rather not change the Flink repo for the benchmark needs, WDYT?

masteryhx: Fine. I not insisted on it.

masteryhx (commented):
LGTM.

PR #83, opened 2023-12-16
[FLINK-30535] Introduce TTL state based benchmarks

Zakelly:
As the title says, This PR introduce TTL state based benchmarks. The newly added tests are basically the same as the previous tests, but have been streamlined because the running time is too long with so many parameters.

I also suggest we disable it in the daily run, or only enable it weekly in jenkins since this is time-consuming.

### DIFF
--- a/src/main/java/org/apache/flink/state/benchmark/ttl/TtlListStateBenchmark.java
+++ b/src/main/java/org/apache/flink/state/benchmark/ttl/TtlListStateBenchmark.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.state.benchmark.ttl;
+
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend;
+import org.apache.flink.state.benchmark.StateBenchmarkBase;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.VerboseMode;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.apache.flink.state.benchmark.StateBackendBenchmarkUtils.applyToAllKeys;
+import static org.apache.flink.state.benchmark.StateBackendBenchmarkUtils.compactState;
+import static org.apache.flink.state.benchmark.StateBackendBenchmarkUtils.getListState;
+import static org.apache.flink.state.benchmark.StateBenchmarkConstants.listValueCount;
+import static org.apache.flink.state.benchmark.StateBenchmarkConstants.setupKeyCount;
+
+/** Implementation for list state benchmark testing. */
+public class TtlListStateBenchmark extends TtlStateBenchmarkBase {
+    private final String STATE_NAME = "listState";
+    private final ListStateDescriptor<Long> STATE_DESC =
+            configTtl(new ListStateDescriptor<>(STATE_NAME, Long.class));
+    private ListState<Long> listState;
+    private List<Long> dummyLists;
+
+    public static void main(String[] args) throws RunnerException {
+        Options opt =
+                new OptionsBuilder()
+                        .verbosity(VerboseMode.NORMAL)
+                        .include(".*" + TtlListStateBenchmark.class.getCanonicalName() + ".*")
+                        .build();
+
+        new Runner(opt).run();
+    }
+
+    @Setup
+    public void setUp() throws Exception {
+        keyedStateBackend = createKeyedStateBackend();
+        listState = getListState(keyedStateBackend, STATE_DESC);
+        dummyLists = new ArrayList<>(listValueCount);
+        for (int i = 0; i < listValueCount; ++i) {
+            dummyLists.add(random.nextLong());
+        }
+        keyIndex = new AtomicInteger();
+    }
+
+    @Setup(Level.Iteration)
+    public void setUpPerIteration() throws Exception {
+        for (int i = 0; i < setupKeyCount; ++i) {
+            keyedStateBackend.setCurrentKey((long) i);
+            listState.add(random.nextLong());
+        }
+        // make sure only one sst file left, so all get invocation will access this single file,
+        // to prevent the spike caused by different key distribution in multiple sst files,
+        // the more access to the older sst file, the lower throughput will be.
+        if (keyedStateBackend instanceof RocksDBKeyedStateBackend) {
+            RocksDBKeyedStateBackend<Long> rocksDBKeyedStateBackend =
+                    (RocksDBKeyedStateBackend<Long>) keyedStateBackend;
+            compactState(rocksDBKeyedStateBackend, STATE_DESC);
+        }
+    }
+
+    @TearDown(Level.Iteration)
+    public void tearDownPerIteration() throws Exception {
+        applyToAllKeys(
+                keyedStateBackend,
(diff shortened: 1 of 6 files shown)


Zakelly:
@masteryhx @Myasuka  would you please take a look? Many thanks~

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.TimeUnit;
+
+/** The base class for state tests with ttl. */
+public class TtlStateBenchmarkBase extends StateBenchmarkBase {
+
+    /** The expired time of ttl. */
+    public enum ExpiredTimeOptions {
+
+        /** 5 seconds. */
+        Seconds5(5000),
```
Myasuka, line 18: What's the performance/result change if we tune this expired-seconds configuration?
Say if we increase it to `10` seconds, or decrease it to `3` seconds?

src/main/java/org/apache/flink/state/benchmark/ttl/TtlListStateBenchmark.java
```diff
@@ -0,0 +1,150 @@
...
+    @Setup(Level.Iteration)
+    public void setUpPerIteration() throws Exception {
+        for (int i = 0; i < setupKeyCount; ++i) {
+            keyedStateBackend.setCurrentKey((long) i);
+            listState.add(random.nextLong());
+        }
+        // make sure only one sst file left, so all get invocation will access this single file,
+        // to prevent the spike caused by different key distribution in multiple sst files,
+        // the more access to the older sst file, the lower throughput will be.
+        if (keyedStateBackend instanceof RocksDBKeyedStateBackend) {
```
Myasuka, line 84: I noticed that there existed so much duplicated code here, can how reduce this code via Generics or more abstract class?

Myasuka (commented):
Thanks for creating this PR, I mainly care about the choose of TTL configuration.

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
Zakelly: Considering that the warm-up will take `10` seconds, `5` seconds is enough for the test to enter the key expiration phase, and it is not too short. I will give a comparison of the results under different configurations.

src/main/java/org/apache/flink/state/benchmark/ttl/TtlListStateBenchmark.java
Zakelly: Actually, in order to prevent small changes from affecting the existing test results, I did not use class abstraction purposely. I suggest we introduce a refactor to the existing tests along with the [FLINK-33864](https://issues.apache.org/jira/browse/FLINK-33864), since it can greatly affect test results. WDYT?

Myasuka: I'm not sure whether such refactor could impact the performance results significantly, could you please provide some data if possible? If so, I think we can refactor the tests with FLINK-33864.

Zakelly: After some try, I found there is no easy way to do so, since the `Ttl*StateBenchmark` extends the `TtlStateBenchmarkBase`, while they cannot extends `*StateBenchmark` in the meantime. So a possible way is to extract an `TestHarness` or `TestContext` to hold the variables. I only do refactor in `valueAdd`, and test it on my local machine. Before:
```
Benchmark                     (backendType)   Mode  Cnt     Score     Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5815.682 ± 122.008  ops/ms
Benchmark                     (backendType)   Mode  Cnt     Score     Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5817.108 ± 162.588  ops/ms
Benchmark                     (backendType)   Mode  Cnt     Score     Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5792.671 ± 132.356  ops/ms
```
And after the refactor:
```
Benchmark                     (backendType)   Mode  Cnt     Score     Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5712.225 ± 133.996  ops/ms
Benchmark                     (backendType)   Mode  Cnt     Score    Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5733.398 ± 77.725  ops/ms
Benchmark                     (backendType)   Mode  Cnt     Score    Error   Units
ValueStateBenchmark.valueAdd           HEAP  thrpt   30  5771.855 ± 80.658  ops/ms
```
As we can see, there is about 1% impact on the performance results.

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
Zakelly: I think it is better to manipulate the `TtlTimeProvider` to finely control the number of keys eliminated in each iteration. So I customize the `TtlTimeProvider` and provide an option for the percentage of keys that expired per iteration. This option will affect the result of test especially in `valueGet`. Here's a running test:
```
# Benchmark: org.apache.flink.state.benchmark.ttl.TtlValueStateBenchmark.valueGet
# Parameters: (backendType = HEAP, expiredOption = ExpirePercent3PerIteration, stateVisibility = NeverReturnExpired, updateType = OnCreateAndWrite)

# Run progress: 0.00% complete, ETA 00:12:00
# Fork: 1 of 3
# Warmup Iteration   1: 2993.246 ops/ms
# Warmup Iteration   2: 3520.954 ops/ms
# Warmup Iteration   3: 3640.884 ops/ms
# Warmup Iteration   4: 3643.386 ops/ms
# Warmup Iteration   5: 3776.677 ops/ms
# Warmup Iteration   6: 3739.375 ops/ms
# Warmup Iteration   7: 3903.434 ops/ms
# Warmup Iteration   8: 3877.286 ops/ms
# Warmup Iteration   9: 3913.663 ops/ms
# Warmup Iteration  10: 4093.471 ops/ms
Iteration   1: 4117.018 ops/ms
Iteration   2: 4255.166 ops/ms
Iteration   3: 4338.564 ops/ms
Iteration   4: 4439.571 ops/ms
Iteration   5: 4603.235 ops/ms
Iteration   6: 4735.039 ops/ms
Iteration   7: 4896.946 ops/ms
Iteration   8: 5153.644 ops/ms
Iteration   9: 5429.196 ops/ms
Iteration  10: 5772.806 ops/ms
```
As we can see, as the key expires, the test results increase iteration by iteration. I pick 3% for a predefined value, since there are 20 iterations in each test and we cannot let all the keys expire. WDYT? @Myasuka

Myasuka: I like the idea to use expired percent instead of the specific time-to-live. BTW, why choose 3% as the predefined value?

src/main/java/org/apache/flink/state/benchmark/ttl/TtlListStateBenchmark.java
Myasuka: I think the 1% impact is really small compared to the project maintaining. And we can postpone this later in [FLINK-33864](https://issues.apache.org/jira/browse/FLINK-33864).

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
```diff
@@ -3,46 +3,102 @@
...
 
+    private static final long initialTime = 1000000;
+
     /** The expired time of ttl. */
     public enum ExpiredTimeOptions {
 
-        /** 5 seconds. */
-        Seconds5(5000),
+        /** Expire 3 percent of the initial keys per iteration. */
+        ExpirePercent3PerIteration(3),
```
Myasuka, line 24: Should be `Expire3PercentPerIteration(3)`

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
```diff
@@ -3,46 +3,102 @@
...
                         .build();
         stateDescriptor.enableTimeToLive(ttlConfig);
         return stateDescriptor;
     }
+
+    protected KeyedStateBackend<Long> createKeyedStateBackend() throws Exception {
+        KeyedStateBackend<Long> originalBackend = super.createKeyedStateBackend();
+        timeProvider = new ControllableTtlTimeProvider();
+        if (originalBackend instanceof AbstractKeyedStateBackend) {
+            // TODO: Config TtlTimeProvider in proper way.
```
Myasuka, line 61: I think we can introduce a `TtlTimeProvider` in Flink's interface.

Myasuka (commented):
Thanks for the update, I like the idea of letting the fixed data expire better than a configured TTL seconds. Please take a look at my comments.

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
Zakelly: Sure, I make a PR for this https://github.com/apache/flink/pull/23985

src/main/java/org/apache/flink/state/benchmark/ttl/TtlStateBenchmarkBase.java
Zakelly: There are 20 iterations in each test, 10 rounds warm-up and 10 rounds testing. If 3% is chosen, the testing will cover 70% down to 40% state size, which is a relatively reasonable value (not so big or small).

Zakelly:
@Myasuka I fixed a compile error and now it's OK. Maybe this repo needs a CI process validating the compilation.
Please let me know if you have any other concern on this PR. Thanks.

Zakelly:
@Myasuka Kindly remind~

Myasuka (approved):
Thanks for the update, LGTM.

PR #88, opened 2024-02-18
[FLINK-34453] Add warmup in SchedulerBenchmarkExecutorBase

Zakelly:
This will stabilize the perf score of benchmarks for runtime scheduler.


src/main/java/org/apache/flink/scheduler/benchmark/SchedulerBenchmarkExecutorBase.java
```diff
@@ -37,6 +38,7 @@
 @State(Scope.Thread)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @BenchmarkMode(Mode.AverageTime)
+@Warmup(iterations = 20)
```
fredia, line 41: How about also specifying @Measurement?

Zakelly:
I had a discussion with Zhuzhu and found that the scheduler benchmarks is aim to measure the code start time. So no warmup should be added. I'm closing this.

PR #89, opened 2024-05-17
[hotfix] Fix compilation error due to constructor removal of MockSourceReader

Zakelly:
Due to this commit in main repo https://github.com/apache/flink/commit/41391065e1bf1476040039077049fbb98bec457e , there is a compilation error. This PR change to use another constructor of `MockSourceReader` to fix this.


fredia (approved):
Thanks for the quick fix, LGTM

pnowojski:
Thanks for fixing it!

PR #90, opened 2024-05-26
Enable async profiler

SamBarker:
In exploring [FLINK-35215](https://issues.apache.org/jira/browse/FLINK-35215) being able to generate flame graphs from each benchmark run makes comparing results a lot easier.

However specifying the configuration of the profiler using the exec plugin is problematic as we need to conditionally add an argument based on the value of a property. So instead of duplicating the existing profiles to add profiler when the new profiles are activated delegate the execution to a shell script which can handle the optional parameters.


pom.xml
```diff
@@ -345,6 +346,8 @@ under the License.
 								<argument>-classpath</argument>
 								<classpath/>
 								<argument>org.openjdk.jmh.Main</argument>
+								<argument>-prof</argument>
+								<argument>async:libPath=${asyncProfilerLib};output=flamegraph;dir=/tmp/profile-results</argument>
```
SamBarker, line 350: Argh. this fails if `asyncProfilerLib` is invalid.

pom.xml
```diff
@@ -536,6 +543,19 @@ under the License.
...
+		<profile>
+			<activation>
+				<activeByDefault>false</activeByDefault>
+				<property>
+					<name>asyncProfilerLib</name>
+				</property>
+			</activation>
+			<id>enable-async-profiler</id>
+			<properties>
+				<jmhProfArgument>async:libPath=${asyncProfilerLib};output=flamegraph;dir=/tmp/profile-results</jmhProfArgument>
```
SamBarker, line 556: `libPath` is user and platform specific so I don't think we can sensibly default it.

benchmark.sh
```diff
@@ -0,0 +1,54 @@
...
+
+while getopts ":j:c:b:e:p:a:m:h" opt; do
+  case $opt in
+    j) JAVA_ARGS+=("${OPTARG}")
+    ;;
+    c) CLASSPATH_ARG="${OPTARG}"
+    ;;
+    b) BINARY="${OPTARG}"
+    ;;
+    p) PROFILER_ARG="${OPTARG:+-prof ${OPTARG}}"
```
SamBarker, line 16: This is the core motivation for the change. If `-p` is specified with an non empty value this will set `PROFILER_ARG` to `-prof ${OPTARG}`. Without this it would end up adding `-prof " "` and thus JMH fails to initialise.

SamBarker:
I think the `benchmark-java17` profile could potentially be simplified to specify a property with just the additional JVM_ARGS (possibly changing `-j` to be single arg inline with `-p`). 

However I wanted to gauge interest in this change before I went any further with potential refactoring. 

I also think it would be very helpful to capture the flamegraphs as part of the information included in `flink-speed` but I have no idea about the implications of that or the required setup for the build server so haven't perused that option as yet.

benchmark.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
SamBarker, line 1: One final call out. This is not currently checked in with the executable bit set as I'm not sure what the projects preference would be for that.

pom.xml
```diff
@@ -54,11 +54,11 @@ under the License.
 		<maven.compiler.target>${java.version}</maven.compiler.target>
 		<slf4j.version>1.7.36</slf4j.version>
 		<log4j.version>1.2.17</log4j.version>
-		<jmh.version>1.19</jmh.version>
+		<jmh.version>1.37</jmh.version>
```
SamBarker, line 57: Needed a newer version for async-profiler support in JMH so went for the most recent release.

pom.xml
```diff
@@ -54,11 +54,11 @@ under the License.
...
 		<junit.version>4.13.2</junit.version>
 		<avro.version>1.11.1</avro.version>
 		<mockito.version>3.4.6</mockito.version>
-		<maven.exec.version>1.6.0</maven.exec.version>
+		<maven.exec.version>3.3.0</maven.exec.version>
```
SamBarker, line 61: The old version had issues trying to log `-p` so went for the most recent release.

benchmark.sh
```diff
@@ -0,0 +1,48 @@
...
+usage: TODO
+EOF
+      exit 1
+    ;;
+    \?) echo "Invalid option -$opt ${OPTARG}" >&2
+    exit 1
+    ;;
+  esac
+done
+shift "$(($OPTIND -1))"
```
pnowojski: I would suggest to use python or java wrapper instead of bash, if only for the sake of being able to use a nice args library. For example take a look at https://github.com/apache/flink-benchmarks/blob/master/regression_report.py or some other python script in this repo's root.

pom.xml
```diff
@@ -290,25 +291,31 @@ under the License.
 						<configuration>
 							<skip>${skipTests}</skip>
 							<classpathScope>test</classpathScope>
-							<executable>${executableJava}</executable>
+							<executable>${basedir}/benchmark.sh</executable>
```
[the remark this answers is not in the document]
pnowojski: Are the executable commands the same? There are some scripts (in [jenkins](https://github.com/apache/flink-benchmarks/tree/master/jenkinsfiles)?) that are relaying for those commands to behave the way they were behaving. If something changes, they would have to be changed as well.

pnowojski (commented):
Thanks for the contribution. 

AFAIU, by default profiling is not turned on. One has to activate `enable-async-profiler` profile?

Would be nice if someone (@Zakelly?) who controls how the benchmarks are being executed right now (they used to be triggered from a jenkins machine, using jenkins files checked in this repo) reviewed this as well.

pom.xml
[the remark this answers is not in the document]
SamBarker: The interface with maven is identical so how the Java env is controlled stays consistent. The change is in how Jenkins triggers the benchmarks.

SamBarker:
> AFAIU, by default profiling is not turned on. One has to activate enable-async-profiler profile?

Yes that's almost right. The profile is enabled automatically by specifying the path the profiler shared library (as opposed to running `-P enable-async-profiler`) Async profiler needs downloaded manually and the path is platform specific I didn't want to tackle automation of that unless there was some interest in the idea.

Zakelly:
Thanks for the contribution! I'll try this out in our `jenkins` env.

benchmark.sh
SamBarker: I had a go at doing it in python https://github.com/SamBarker/flink-benchmarks/blob/generate-flamegraphs-py/run-benchmarks.py life got quite funky trying to pass `--add-opens` to something that actually understood long arguments. 

I can merge that into this PR if preferred but I'm not sure it actually improves things here. 

The other option might be to look at [ap-loader](https://github.com/jvm-profiling-tools/ap-loader) which would allow us to ensure that the native libraries are available.

SamBarker:
@Zakelly any news on the jenkins env? I'm largely hoping its a no-op there right now.

@pnowojski readme and executable bit set. Is there anything else your looking for?

README.md
```diff
@@ -64,6 +64,20 @@ java -jar target/benchmarks.jar "<benchmark_class>" -lp
 java -jar target/benchmarks.jar "org.apache.flink.state.benchmark.*" -p "backendType=ROCKSDB" 
 ```
 
+## Generating Flame graphs
+
+JMH has support for enabling profilers when running benchmarks, in particular [async profiler](https://github.com/async-profiler/async-profiler) which can generate flame graphs of the call stack. However, async profiler requires linking with native code which needs downloaded manually which means its location is user and architecture specific. To enable async profiler support run the benchmarks as follows:
+
+with maven
+```
+ java -jar target/benchmarks.jar -rf csv "<benchmark_class>" -DasyncProfilerLib=<PATH_TO_libasyncProfiler.*>
```
showuon, line 73: nit: additional space at the beginning.

showuon (commented):
Nice improvement! Thanks!

pom.xml
```diff
@@ -67,6 +67,7 @@ under the License.
 		<benchmarkExcludes>org.apache.flink.benchmark.full.*,org.apache.flink.state.benchmark.*,org.apache.flink.scheduler.benchmark.*</benchmarkExcludes>
 		<benchmarks>.*</benchmarks>
 		<executableJava>java</executableJava>
+		<jmhProfArgument> </jmhProfArgument>
```
Zakelly, line 70: How about adding a property for path for profiling result? And I suggest a relative path under current path for default value.

benchmark.sh
Zakelly, line 1: `chown +x` to make this executable?

Zakelly (commented):
I tried in current jenkins env and it runs well. I'm still debugging the scripts to make jenkins archive all the profiling results. Despite of that, I have some minor comments as below:

Zakelly:
And I find the benchmark runs extremely slow with the profiler on. The ETA is up to 8 days for full set of our daily run. So I guess this is best only used for single tests that are triggered manually, right?

benchmark.sh
SamBarker: resolved with https://github.com/apache/flink-benchmarks/pull/90/commits/74e4f4f56dab07c70e03690279f96a8f168830f4

SamBarker:
> And I find the benchmark runs extremely slow with the profiler on. The ETA is up to 8 days for full set of our daily run. So I guess this is best only used for single tests that are triggered manually, right?

Yikes thats a bit sad. Running a single test (`SerializationFrameworkMiniBenchmarks.serializerHeavyString`) locally I get 10 mins estimate with or without the profiler and a runtime that seems comparable. If I widen the set of benchmarks I get the same ETA with or without the profiler. 

So I'm slightly suspicious that the excludes or default includes behaviour is being changed but without knowing what command line your seeing that from its hard to know.

Zakelly:
Here's the full log before the pipeline expired. [jenkins.tar.gz](https://github.com/apache/flink-benchmarks/files/15484257/jenkins.tar.gz)

It compiles & install flink before run benchmarks via `/mnt/jenkins/tools/hudson.tasks.Maven_MavenInstallation/M3/bin/mvn -Dflink.version=1.20-SNAPSHOT clean install exec:exec -Penable-async-profiler -DskipTests -Drat.skip=true '-Dbenchmarks=org.apache.flink.state.benchmark.*' '-DbenchmarkExcludes=org.apache.flink.benchmark.full.*' -DexecutableJava=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.372.b07-4.0.2.al8.x86_64/bin/java -DasyncProfilerLib=/opt/async-profiler-3.0-linux-x64/lib/libasyncProfiler.so -DprofResultPath=./profile-result`


It would be great if you could provide some options that could help speed this up.

SamBarker:
> It compiles & install flink before run benchmarks via /mnt/jenkins/tools/hudson.tasks.Maven_MavenInstallation/M3/bin/mvn -Dflink.version=1.20-SNAPSHOT clean install exec:exec -Penable-async-profiler -DskipTests -Drat.skip=true '-Dbenchmarks=org.apache.flink.state.benchmark.*' '-DbenchmarkExcludes=org.apache.flink.benchmark.full.*' -DexecutableJava=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.372.b07-4.0.2.al8.x86_64/bin/java -DasyncProfilerLib=/opt/async-profiler-3.0-linux-x64/lib/libasyncProfiler.so -DprofResultPath=./profile-result
>It would be great if you could provide some options to speed this up.

It looks to me like it was handling the excludes and includes correctly which is what I was concerned about. I can't reproduce a slower runtime with the profiler on
With profiler
```
mvn -Dflink.version=1.20-SNAPSHOT -DexecutableJava=${JAVA_HOME}/bin/java -DasyncProfilerLib=/home/sam/src/async-profiler/async-profiler-3.0-linux-x64/lib/libasyncProfiler.so -Dbenchmarks="SerializationFrameworkMiniBenchmarks.serializerHeavyString" exec:exec --offline
... 
snip 
...

# JMH version: 1.37
# VM version: JDK 11.0.23, OpenJDK 64-Bit Server VM, 11.0.23+9
# VM invoker: /home/sam/.sdkman/candidates/java/11.0.23-tem/bin/java
# VM options: -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.ssl
# Blackhole mode: full + dont-inline hint (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 10 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.apache.flink.benchmark.SerializationFrameworkMiniBenchmarks.serializerHeavyString

# Run progress: 0.00% complete, ETA 00:10:00
... 
snip 
...
 [INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  11:17 min
[INFO] Finished at: 2024-05-30T12:10:42+12:00
[INFO] ------------------------------------------------------------------------

```

without profiler
```
mvn -Dflink.version=1.20-SNAPSHOT -DexecutableJava=${JAVA_HOME}/bin/java -Dbenchmarks="SerializationFrameworkMiniBenchmarks.serializerHeavyString" exec:exec --offline
... 
snip 
...
# JMH version: 1.37
# VM version: JDK 11.0.23, OpenJDK 64-Bit Server VM, 11.0.23+9
# VM invoker: /home/sam/.sdkman/candidates/java/11.0.23-tem/bin/java
# VM options: -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.ssl
# Blackhole mode: full + dont-inline hint (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 10 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.apache.flink.benchmark.SerializationFrameworkMiniBenchmarks.serializerHeavyString

# Run progress: 0.00% complete, ETA 00:10:00
... 
snip
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  11:18 min
[INFO] Finished at: 2024-05-30T12:27:06+12:00
[INFO] ------------------------------------------------------------------------
```

~So I'm not really sure what to suggest.~

Wait a minute. This PR changes the version of JMH. Its not the profiler being ON or OFF that affects the runtime its the JMH version.

going back to main I get 

```
# JMH version: 1.19
# VM version: JDK 11.0.23, VM 11.0.23+9
# VM invoker: /home/sam/.sdkman/candidates/java/11.0.23-tem/bin/java
# VM options: -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.ssl
# Warmup: 10 iterations, 1 s each
# Measurement: 10 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.apache.flink.benchmark.SerializationFrameworkMiniBenchmarks.serializerHeavyString

# Run progress: 0.00% complete, ETA 00:01:00
```

This may be related to changes in warmups in [JMH 1.21](https://mail.openjdk.org/pipermail/jmh-dev/2018-May/002753.html) 
> *) Defaults for warmup/measurement/forks were reconsidered to allow longer in-JVM runs. This is
important for the VMs/benchmarks where time-to-performance is substantial. Instead of 20x1s warmup,
20x1s measurement, 10 forks, we are now doing 5x10s warmup, 5x10s measurement, 5 forks. It increases
the default time for around 1.25x, and single fork has 2.5x more time to warmup and complete.

SamBarker:
I've confirmed the runtime duration increase comes from upgrading JMH from 1.19 to 1.21. The minimum version for async profiler looks to be JMH 1.31. So if we want flame graphs 1.37 still feels like the right choice. 

What I haven't been able to determine is what the change in warmup behaviour implies for the insights of the current benchmark suite and how comparable through time the results would be.

SamBarker:
@Zakelly I've pushed https://github.com/apache/flink-benchmarks/pull/90/commits/decdc8e2d161584a798461b4f381adba7cee731f which restores the JMH defaults from 1.19. Hopefully that addresses the issues with the runtime under Jenkins. I'd just as happily revert that commit but it de-couples this PR from reasoning about the implications for changing the warmup and measurement times.

Zakelly:
> @Zakelly I've pushed [decdc8e](https://github.com/apache/flink-benchmarks/commit/decdc8e2d161584a798461b4f381adba7cee731f) which restores the JMH defaults from 1.19. Hopefully that addresses the issues with the runtime under Jenkins. I'd just as happily revert that commit but it de-couples this PR from reasoning about the implications for changing the warmup and measurement times.

I see. Thanks for your investigation. I'll give it a try.

SamBarker:
@Zakelly @pnowojski do you need anything else from me to progress this? Would one of you be able to approve this PR to run through the workflow so I can verify its good to go?

Zakelly (approved):

PR #95, opened 2024-09-18
[hotfix] Use `Duration` instead of `Time`

Zakelly:
Due to https://github.com/apache/flink/pull/25250 , the TTL builder using `org.apache.flink.api.common.time.Time` is removed. This PR fixes this.


Zakelly:
@XComp would you please take a look?

XComp (approved):
Thanks for fixing it. `org.apache.flink.streaming.api.windowing.time.Time` isn't merged, yet. But it makes sense to fix everything in one go. 👍 

I verified that there are no other occurrences of `Time` in the code base using the following regex: `[^a-z]Time[^A-Za-z]`

Zakelly:
> Thanks for fixing it. `org.apache.flink.streaming.api.windowing.time.Time` isn't merged, yet. But it makes sense to fix everything in one go. 👍
> 
> I verified that there are no other occurrences of `Time` in the code base using the following regex: `[^a-z]Time[^A-Za-z]`

Thanks for the double check.

PR #97, opened 2024-10-09
[FLINK-36322] Fix compile errors based on 2.0-preview

Zakelly:
There are many breaking changes in flink 2.0. This PR fixes all the compile error based on the 2.0-preview.

Note: Some tests are removed since the corresponding features/options are removed in the main repo.

TODO: Replace usage of `SinkFunction` and `RichParallelSourceFunction` before 2.0 release.


fredia (approved):
Thanks for the PR, LGTM

PR #99, opened 2025-01-14
[FLINK-37115] Fix compile error related with SerializerConfig

X-czh:
A recent breaking change in Flink 2.0 (https://github.com/apache/flink/pull/25960) breaks compilation. This PR aims to fix that.


Zakelly (approved):
Thanks for the fix! LGTM

PR #103, opened 2025-01-23
[hotfix] Add profile to apply `--add-opens` for java17 CI

Zakelly:


Zakelly:
As we can see, the `--add-opens` is successfully added during CI testing. 
image
https://github.com/apache/flink-benchmarks/actions/runs/12923911676/job/36041994892 (CI based on flink-2.0 since in 2.1 we still have compile issue).

reswqa (approved):

PR #106, opened 2025-08-06
[FLINK-38199] Add SinkUpsertMaterializerBenchmark

rkhachatryan:
Requires https://github.com/apache/flink/pull/26876

### DIFF
--- a/src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
+++ b/src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
@@ -132,7 +132,7 @@ public class SinkUpsertMaterializerBenchmark extends BenchmarkBase {
         public int numRecordsTotal;
 
         // larger payload amplifies any inefficiencies but slows down the benchmark; mostly affects rocksdb
-        @Param({"10", "250"})
+        @Param({"20"})
         public int payloadSize;
 
         // lower retraction percentage implies longer history, making retractions even harder (unless percentage = 0)


pom.xml
```diff
@@ -45,7 +45,7 @@ under the License.
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<flink.version>2.2-SNAPSHOT</flink.version>
+		<flink.version>2.1-SNAPSHOT</flink.version>
```
pnowojski, line 48: 2.2 -> 2.1?

src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
```diff
@@ -0,0 +1,298 @@
...
+
+    public static void main(String[] args) throws RunnerException {
+        new Runner(new OptionsBuilder()
+                .verbosity(VerboseMode.NORMAL)
+                // speedup
+//                .warmupIterations(1)
+//                .measurementIterations(1)
+//                .forks(1)
+//                .warmupTime(TimeValue.milliseconds(100))
+//                .measurementTime(TimeValue.seconds(1))
```
pnowojski, line 113: left overs?

src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
```diff
@@ -0,0 +1,298 @@
...
+
+        @Param({"false", "true"}) public boolean hasUpsertKey;
+
+        @Param({"HEAP", "ROCKSDB"}) public SumStateBackend stateBackend;
+
+        public int numRecordsTotal;
+
+        // larger payload amplifies any inefficiencies but slows down the benchmark; mostly affects rocksdb
+        @Param("100")
+        public int payloadSize;
```
pnowojski, line 142: For micro benchmark purposes, like optimising code, guarding against regressions, the lower the payload, the better, as you get more contrasting effects of any performance related changes - optimisations/changes to the related code will be more clearly visible instead of being watered down by bytes copy.

So I would shrink it down, maybe to 10 or 20.

src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
```diff
@@ -0,0 +1,298 @@
...
+
+        // lower retraction percentage implies longer history, making retractions even harder (unless percentage = 0)
+        @Param("100")
+        public int retractPercentage;
+
+        // higher retraction delay leaves longer history, making retractions even harder (unless percentage = 0)
+        // for automated runs, reduce the run time (and the data points) to the most common cases
+        @Param({"1", "100"})
```
pnowojski, line 150: 1 and 1000 for automated runs?

pnowojski (changes requested):

src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
rkhachatryan: I think both cases can be problematic; I'll change to a smaller value and think about keeping 100 (or increasing it).

src/main/java/org/apache/flink/benchmark/SinkUpsertMaterializerBenchmark.java
rkhachatryan: I wanted to keep it to quickly test changes locally, but I can remove

pnowojski: Please remove. You can also easily change that in the `BenchmarkBase`. (I almost aloways have have locally unocommited changes to `BenchmarkBase.java`)

pnowojski (commented):
mostly LGTM % opened comments

PR #117, opened 2026-05-28
[FLINK-39787][build] Map runner hostname to loopback to avoid SSL SAN mismatch

MartijnVisser:
GitHub-hosted Azure runners assign a VM hostname whose reverse DNS returns an internal cloudapp.net FQDN. flink-runtime's SSL test certificate (local127.keystore) lists only 'localhost', '127.0.0.1', and '::1' as SANs, so the SSL handshake in
StreamNetworkThroughputBenchmark fails and the benchmark hangs until GitHub kills the job at the 6h timeout.

Prepend a loopback mapping for the runner hostname to /etc/hosts so InetAddress.getLocalHost() (used by StreamNetworkBenchmarkEnvironment to bind the NettyServer) returns 127.0.0.1; reverse DNS on that IP then yields 'localhost', which matches the certificate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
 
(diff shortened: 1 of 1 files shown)


alpinegizmo (approved):
LGTM

